PR #55, opened 2024-12-10
Add option to write to a file during solve

ilateur:
Pretty self-explanatory.
Should there also be an option for extremely large simulations to only write to a table, without keeping data in a dictionary in RAM?

### DIFF
--- a/mumaxplus/timesolver.py
+++ b/mumaxplus/timesolver.py
@@ -24,7 +24,8 @@ class TimeSolverOutput:
         self._keys = list(self._data.keys())  # keep list of keys to maintain order
         self._file_name = file_name
         if self._file_name is not None:
-            _os.makedirs(_os.path.dirname(self._file_name), exist_ok=True)
+            if directory := _os.path.dirname(self._file_name):
+                _os.makedirs(directory, exist_ok=True)
             with open(self._file_name, 'w') as file:  # make new file
                 print("# " + "\t".join(self._keys), file=file)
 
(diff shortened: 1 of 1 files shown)


mumaxplus/timesolver.py
```diff
@@ -9,20 +10,35 @@ class TimeSolverOutput:
...
+    def __init__(self, quantity_dict, file_name=None):
         self._quantities = quantity_dict
         self._data = {"time": []}
         for key in self._quantities.keys():
             self._data[key] = []
 
+        self._keys = list(self._data.keys())  # keep list of keys to maintain order
+        self._file_name = file_name
+        if self._file_name is not None:
+            _os.makedirs(_os.path.dirname(self._file_name), exist_ok=True)
```
lamoreel, line 27: Only works if `self._file_name` contains a directory path, so giving only a file name raises a `FileNotFoundError`.

lamoreel (changes requested):
An option to avoid storing all data in RAM might be useful, but I would keep this very optional in order to keep the option to manipulate output in the same input file. Is a new function which only writes necessary? Or might `solve` just have another optional argument?

mumaxplus/timesolver.py
ilateur: Silly

ilateur:
I suggest we add that option later, but merge this feature like it is now

lamoreel:
If it works now, then yes.

PR #53, opened 2024-12-03
Update versions of CMake, Python and pybind11

ilateur:
- Update CMake 3.16 to 3.31.1, with minimum of 3.24 for clean automatic CUDA architecture detection
- Update Python from 3.8 to 3.13
- Update pybind11 from ~v2.4 to v2.13.6
- Set CMAKE_INTERPROCEDURAL_OPTIMIZATION to OFF to disable pybind::(thin_)lto
- Fix mumax3 Slonczewski test after Python (NumPy) update

This changes both the conda environment and the pybind11 submodule. Please test out if everything works on your machines too before merging. After checking out this branch, update the conda environment with
```bash
conda env update --file environment.yml
```
and update pybind11 with
```bash
git update submodule
```
Afterwards (or maybe even before), git might keep bothering you about uncommitted pybind11 changes. You might need to go into the `src/bindings/pybind11/` directory, run `git status` yourself and remove unwanted directories or undo unwanted changes. (A totally new user will not need to do this.)

PR #56, opened 2024-12-18
Fixed pybind11 bug

hernantech:
- pybind11 v2.5 has a bug which shows during compilation 
`error: 'uint16_t' in namespace 'std' does not name a type; did you mean 'wint_t'?`
- README's `clone --recursive` currently points to v2.5
- Easier to just force any version greater than v2.6 rather than rewriting files with `uint16_t` references to a compatible type

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,4 @@
 [submodule "src/bindings/pybind11"]
 	path = src/bindings/pybind11
 	url = https://github.com/pybind/pybind11
+	branch = v2.6.2


ilateur:
This bug does not seem to occur with older versions of gcc (below v13?). I had also tried to update pybind11 before, but ran into different issues. Is v2.6.2 fine?

Regardless, I think the cleanest solution is to update everything (pybind11, Python and CMake) together, as is done in pull request #53. It should eventually make its way to the master branch after everyone recovers from the holidays :)

PR #57, opened 2025-01-15
Add proper adaptive timestepping to elastic simulations

ilateur:
# Problem
Adaptive time stepping did not properly work with magnetoelastic simulations, because the absolute error of all Variables was used to gauge accuracy. An absolute error of 1e-5 or lower was seen as good enough for displacements of the order of ~1e-14 m and velocities of ~1e-4 m/s.

# Solution
The essential difference is moving the maxError of adaptive time stepping from TimeSolver to DynamicEquation, so each equation has its own error to adhere to. This is not user-accessible, so the 3 relevant max errors can be set via the (Ferro)Magnet, with automatic telepathic pointer-communication (I'm not the biggest fan of this solution). This is the only (or maybe not) solver setting outside of TimeSolver :( but it also has a physical unit and is physics dependent.

As a consequence, the Relaxer cannot directly set maxError of the TimeSolver anymore. This is not a problem, because it needs to create new DynamicEquations anyway, so it can also assign a pointer to its own maxerr variable, then scale that.

Let me know if you have a better solution though.

# Potential To-Do's
- take a look at sensibleTimeStep()
- may need to tweak default values 1e-18 and 1e-7 of displacementMaxError and velocityMaxError.
- make starting value of `maxerr_` in Relaxer user-assignable? This used to start at maxError of TimeSolver.
- clean up Relaxer?
- clean up raw pointer logic?

### DIFF
--- a/examples/magnetoelastic.py
+++ b/examples/magnetoelastic.py
@@ -72,9 +72,6 @@ fig, ax = plt.subplots()
 u_scale = 5e4  # amplification of displacement
 u_skip  = 5  # don't show every displacement
 
-world.timesolver.adaptive_timestep = False
-world.timesolver.timestep = 1e-12
-
 steps = 400
 time_max = 0.8e-9
 duration = time_max/steps
--- a/examples/pure_elastic.py
+++ b/examples/pure_elastic.py
@@ -74,10 +74,6 @@ for i in range(3):
     u[i, ...] -= u_avg[i]
 magnet.elastic_displacement = u
 
-# adaptive time stepping does not work for magnetoelastics
-world.timesolver.adaptive_timestep = False
-world.timesolver.timestep = 1e-13
-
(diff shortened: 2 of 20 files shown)


DiegoDeGusem:
The magnetization max error is a ferromagnet property, so it is possible to set 2 different errors in an antiferromagnet? That's a bit strange :thinking: 
Maybe this could also benefit from some tests and/or an example?
I don't immediately see a better way of handling this, so this method is fine by me I think

ilateur:
Do you have a suggestion for any test?
As examples I figured it was fine to remove the fixed time step from all (magneto)elastic examples.

ilateur:
The implementation has been done in a different way, see pull request #60.

ilateur:
The implementation has been done in a different way, see pull request #60.

PR #58, opened 2025-01-15
Add the possibility to set a strain

DiegoDeGusem:
This code allows user to set a normal and/or shear strain. However, it is impossible to combine this with elastodynamics, because elastodynamics wants to modify strain based on displacement. If someone would try this, they get an error.
This pull request also includes a test where the magnetoelastic field from mumax⁺ is compared with mumax³ (the results are exactly the same up to floating points :sunglasses: ).

### DIFF
--- a/mumaxplus/magnet.py
+++ b/mumaxplus/magnet.py
@@ -12,7 +12,7 @@ from .scalarquantity import ScalarQuantity
 from .strayfield import StrayField
 from .variable import Variable
 
-
+import warnings
 class Magnet(ABC):
     """A Magnet should never be initialized by the user. It contains no physics.
     Use ``Ferromagnet`` or ``Antiferromagnet`` instead.
@@ -222,6 +222,10 @@ class Magnet(ABC):
     @enable_elastodynamics.setter
     def enable_elastodynamics(self, value):
         self._impl.enable_elastodynamics = value
+        if _np.any(self.rigid_norm_strain.eval() != 0):
+            raise Exception("You can not use normalized strain with elastodynamics.")
+        elif _np.any(self.rigid_shear_strain.eval() != 0):
+            raise Exception("You can not use shear strain with elastodynamics.")
 
     # ----- ELASTIC PARAMETERS -------
 
@@ -322,6 +326,8 @@ class Magnet(ABC):
     
     @rigid_norm_strain.setter
     def rigid_norm_strain(self, value):
+        if self.enable_elastodynamics:
+            raise Exception("You can not use elastodynamics with normal strain.")
         self.rigid_norm_strain.set(value)
     
     @property
@@ -343,6 +349,8 @@ class Magnet(ABC):
 
     @rigid_shear_strain.setter
     def rigid_shear_strain(self, value):
+        if self.enable_elastodynamics:
+            raise Exception("You can not use elastodynamics with shear strain.")
         self.rigid_shear_strain.set(value)
 
     # ----- ELASTIC QUANTITIES -------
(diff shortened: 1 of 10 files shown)


test/mumax3/test_mumax3_rigid_strain.py
ilateur, line 1: nice, it might be a bit faster to reuse the mumax+ world and magnet across different tests though.

test/mumax3/test_mumax3_rigid_strain.py
```diff
@@ -0,0 +1,110 @@
...
+            exx = {exx}
+            eyy = {eyy}
+            ezz = {ezz}
+            exy = {exy}
+            exz = {exz}
+            eyz = {eyz}
+            B1 = {B}
+            B2 = {B}
+
+            m = Uniform{tuple(magnetization)}
```
ilateur, line 39: Is this arbitrary enough? Maybe safer to use random?

test/mumax3/test_mumax3_rigid_strain.py
```diff
@@ -0,0 +1,110 @@
...
+        err = max_relative_error(result, wanted)
+        assert err < RTOL
+
+    def test_eyz(self):
+        exx, eyy, ezz, exy, exz, eyz = 0, 0, 0, 0, 0, strain
+        magnet, mumax3sim = simulation(exx, eyy, ezz, exy, exz, eyz)
+        wanted = mumax3sim.get_field("B")
+        result = magnet.magnetoelastic_field.eval()
+        err = max_relative_error(result, wanted)
+        assert err < RTOL
```
ilateur, line 110: Maybe also a test with combined (and negative?) strains?

src/physics/magnetoelasticfield.hpp
```diff
@@ -8,6 +8,7 @@ class Field;
 
 // Assure that the magnetoelastic field and force are 0
 bool magnetoelasticAssuredZero(const Ferromagnet*);
+bool appliedStrain(const Magnet*);
```
ilateur, line 11: I would call it `rigidStrainAssuredZero` (and return the opposite), in line with all other assuredZero functions.

ilateur (changes requested):

mumaxplus/magnet.py
```diff
@@ -303,6 +308,52 @@ def rho(self):
...
+        See Also
+        --------
+        elastic_energy, elastic_energy_density, elastic_displacement, stress_tensor
+        """
+        return Parameter(self._impl.rigid_norm_strain)
+    
+    @rigid_norm_strain.setter
+    def rigid_norm_strain(self, value):
+        if self.enable_elastodynamics:
+            raise Exception("You can not use elastodynamics with normal strain.")
```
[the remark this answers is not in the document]
DiegoDeGusem: What do you mean?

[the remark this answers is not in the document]
ilateur: "You can not use normal strain with elastodynamics", because it's a normal strain error

test/mumax3/test_mumax3_rigid_strain.py
ilateur: See `test_parameter` for example for a nice useage of `@pytest.fixture`

src/physics/magnetoelasticfield.cu
```diff
@@ -51,26 +55,65 @@ __global__ void k_magnetoelasticField(CuField hField,
   }
 }
 
+__global__ void k_rigidMagnetoelasticField(CuField hField,
+                                           const CuField mField,
+                                           const CuField normStrain,
+                                           const CuField shearStrain,
```
ilateur, line 61: They should be CuVectorParameters

ilateur (approved):

PR #59, opened 2025-01-16
Add option to use callable coordinate function for time term mask

ilateur:


DiegoDeGusem (approved):
It works, I also tested it with a rigid strain example :D

PR #60, opened 2025-01-20
Add adaptive time stepping to elastic sims via enum in Variable

ilateur:
This is a redo of pull request #57. The idea is exactly the same, but the implementation is different. It was easiest to start over.
Instead of using raw pointers pointing to parameters located in magnet, it uses `enum class MaxError` in `Variable` to remember what error to use. The actual values of the 3 errors are now stored in the TimeSolver class, along with all other adaptive time stepping parameters. I think/hope this solution is cleaner.

Still need to take a look at `sensibleTimeStep()`.


DiegoDeGusem (approved):

ilateur:
Some todo's I should not forget:
- take a look at sensibleTimeStep()
- add a test comparing results with a very small step size
- may need to tweak default values 1e-18 and 1e-7 of displacementMaxError and velocityMaxError

mumaxplus/timesolver.py
```diff
@@ -159,23 +159,65 @@ def time(self, time):
         self._impl.time = time
 
     @property
-    def max_error(self):
-        """Return the maximum error per step the solver can tollerate.
+    def magnetization_max_error(self):
+        """Return the maximum error per step the solver can tollerate for the
```
lamoreel, line 163: Typo: tolerate (also the two extra ones).

lamoreel (changes requested):
Definitely cleaner than the previous fix and I like this for now :+1:. Still need to look at your TODOs before merging.

examples/magnetoelastic_dispersion.py
```diff
@@ -81,10 +81,6 @@ def simulation(theta):
     magnet.alpha = 0
     magnet.eta = 0
 
-    # time stepping
-    world.timesolver.adaptive_timestep = False
-    world.timesolver.timestep = 1e-13
```
ilateur, line 86: I think dispersion relations generally show up better with fixed timestepping 🤔

mumaxplus/timesolver.py
```diff
@@ -210,24 +210,66 @@ def time(self, time):
...
+    @property
+    def displacement_max_error(self) -> float:
+        """Return the maximum error per step the solver can tolerate for the
+        displacement-velocity equations of motion (m).
+        
+        The default value is 1e-18.
+
+        See Also
+        --------
+        headroom, lower_bound, sensible_factor, upper_bound
```
ilateur: also add `sensible_timestep, sensible_timestep_default` like in `magnetization_max_error`

ilateur: `sensible_factor` is specifically designed for torque I think, so it does not make much sense here

mumaxplus/timesolver.py
```diff
@@ -210,24 +210,66 @@ def time(self, time):
...
+    def velocity_max_error(self) -> float:
+        """Return the maximum error per step the solver can tolerate for the
+        velocity-acceleration equations of motion (m/s).
+        
+        The default value is 1e-7.
+
+        See Also
+        --------
+        headroom, lower_bound, sensible_factor, upper_bound
+        displacement_max_error, magnetization_max_error
```
ilateur: same comments as with `displacement_max_error`

mumaxplus/timesolver.py
```diff
@@ -238,6 +280,7 @@ def headroom(self) -> float:
         --------
         lower_bound, max_error, sensible_factor, sensible_timestep,
```
ilateur: `max_error` does not exist anymore (this is mentioned in a lot of docstrings)

mumaxplus/timesolver.py
ilateur: (`sensible_factor` comment remaining)

examples/magnetoelastic_dispersion.py
jplauzie: If you want to keep the adaptive timestepping for performance, you can try interpolating the steps to equidistant points, like how mumax3-fft does? It helps a lot, but is probably not quite as good as a fixed timestep though

PR #61, opened 2025-01-21
Add a rigid strain example

DiegoDeGusem:
This example is based on the simulation and experiment they did in [Mechanism to transfer linear momentum from a Surface Acoustic Wave to a Magnetic Domain Wall](https://arxiv.org/abs/2406.12778).
Its a ferromagnet with a domain wall and by applying a time and space dependent strain (representing a SAW) coming from the right, the wall moves towards the left.
The example shows the initial and final magnetization and also shows a plot of the DW position (estimated by a linear interpolation using the average magnetization) in function of time.


examples/rigid_strain.py
```diff
@@ -0,0 +1,162 @@
...
+# ezz = -E sin(kx - wt)
+# exy = 0
+# exz = E cos(kx - wt)
+# eyz = 0
+# These can all be rewritten by using sum and difference formulas
+# By using add_time_term we can create a f(x,y,z,t) = h(x,y,z)*g(t) function
+# By then adding a second time function we can recreate the effect of the sin and cos
+
+# Splitting everything in single components for the strain results in 8 functions
+def norm_strain_xyz_sin(x,y,z):
```
lamoreel, line 62: Surely these 8 functions can be written out more optimal, perhaps passing a `np.sin` or `np.cos` function as an argument?

examples/rigid_strain.py
```diff
@@ -0,0 +1,162 @@
...
+alpha = 0.02
+K = 7e3
+
+# Elastic coupling constant
+B = -84e6
+B1 = B
+B2 = B
+
+# amplitude, frequency and wave vector of the strain
+E, w, k = 105e-6, 1.3e9 * 2*np.pi, 2*np.pi / 2.8e-6
```
lamoreel, line 27: I would set these at 3 separate lines to improve readability (products and such). Also, why choose `w` for frequency?

examples/rigid_strain.py
```diff
@@ -0,0 +1,162 @@
...
+
+# simulation grid parameters
+nx, ny, nz = 2000, 250, 1
+cx, cy, cz = 4e-9, 4e-9, 3e-9
+
+# create a world and a magnet
+cellsize = (cx, cy, cz)
+grid = Grid((nx, ny, nz))
+world = World(cellsize)
+magnet = Ferromagnet(world, grid) #, geometry=shape.Ellipse(nx*cx, ny*cy).translate(nx*cx/2 - cx/2, ny*cy/2 - cy/2, 0))
```
lamoreel, line 37: Clean-up?

examples/rigid_strain.py
```diff
@@ -0,0 +1,162 @@
...
+magnet.enable_demag = True
```
lamoreel, line 38: This is default.

examples/rigid_strain.py
```diff
@@ -0,0 +1,162 @@
...
+
+for i in range(steps):
+    world.timesolver.run(dt)
+    DW_pos[i] = DW_position(magnet)
+
+# final magnetization
+ax2.imshow(np.transpose(magnet.magnetization.get_rgb()[:,0,:,:], axes=(1,2,0)), origin="lower", extent=im_extent, aspect="equal")
+ax2.set_title("Final magnetization")
+ax2.set_xlabel("x (µm)")
+ax2.set_ylabel("y (µm)")
```
lamoreel, line 154: Double setting.

lamoreel (changes requested):

examples/rigid_strain.py
DiegoDeGusem: How do you give an optional parameter to the mask and add_time_term functions? I thought they should just take x, y and z. You could solve this by doing something like
```Python
def shear_strain_t(t):
    exy = 0
    exz = h(w*t)
    eyz = 0
    return (exy, exz, eyz)

h = np.sin
magnet.rigid_norm_strain.add_time_term(shear_strain_t)
h = np.cos
magnet.rigid_norm_strain.add_time_term(shear_strain_t)
```
but this is also wrong because the first `time_term` will also become a `np.cos` at the next function evaluation.

ilateur: What about something like this:
```python
def strain_t(func):
    return lambda t: (0, func(w*t), 0)

magnet.rigid_norm_strain.add_time_term(strain_t(np.cos))
magnet.rigid_norm_strain.add_time_term(strain_t(np.sin))
```

examples/rigid_strain.py
```diff
@@ -0,0 +1,159 @@
...
+# magnet parameters
+msat = 1.5e6
+aex = 10.5e-12
+alpha = 0.02
+K = 7e3
+
+# Elastic coupling constant
+B = -84e6
+B1 = B
+B2 = B
```
ilateur, line 24: Why not set all these parameters directly to the magnet?

examples/rigid_strain.py
```diff
@@ -0,0 +1,159 @@
+"""In this example we move a domain wall by setting time and space dependant
```
ilateur, line 1: dependent
> a dependant is a person, whereas dependent is an adjective describing a status

examples/rigid_strain.py
ilateur: I'll try to shorten it, I need a doable break from explosions

examples/rigid_strain.py
```diff
@@ -0,0 +1,159 @@
...
+ax1.imshow(np.transpose(magnet.magnetization.get_rgb()[:,0,:,:], axes=(1,2,0)), origin="lower", extent=im_extent, aspect="equal")
+ax1.set_title("Initial magnetization")
+ax2.set_title("Final magnetization")
+ax2.set_xlabel("x (µm)")
+ax2.set_ylabel("y (µm)")
+
+# function to estimate the position of the DW
+def DW_position(magnet):
+    m_av = magnet.magnetization.average()[0]
+    return m_av*nx*cx / 2
```
ilateur, line 141: I wonder if this is exactly correct, given the strange left and right boundaries (`im_extent`) of a magnet :thinking: 
Not that it matters that much in this context.

examples/rigid_strain.py
```diff
@@ -0,0 +1,159 @@
...
+magnet.rigid_norm_strain.add_time_term(norm_strain_t_sin, mask=norm_strain_xyz_cos)
+
+# shear strain
+# Create the first time term f(x,y,z) = h(x,y,z)*g(t)
+magnet.rigid_shear_strain.add_time_term(shear_strain_t_cos, mask=shear_strain_xyz_cos)
+# Add a second time term to obtain sin and cos
+magnet.rigid_shear_strain.add_time_term(shear_strain_t_sin, mask=shear_strain_xyz_sin)
+
+# Create a domain wall
+magnet.magnetization = twodomain((1,0,0), (0,-1,0), (-1,0,0), nx*cx/2, 10*cx)
```
ilateur, line 124: The magnetization should first relax before running, no?

ilateur, line 124: If the domain wall will (mostly) only move to the right, why set it in the middle of the magnet? The left half of the simulation is kinda useless.

examples/rigid_strain.py
```diff
@@ -0,0 +1,159 @@
...
+B2 = B
+
+# amplitude, angular frequency and wave vector of the strain
+E = 105e-6
+w = 1.3e9 * 2*np.pi
+k = 2*np.pi / 2.8e-6
+
+# simulation grid parameters
+nx, ny, nz = 2000, 250, 1
+cx, cy, cz = 4e-9, 4e-9, 3e-9
```
ilateur, line 33: The exchange length of this system is about 2.73e-9, so a cellsize of 4e-9 is fairly large, maybe even too large.
2e-9 would be safer

examples/rigid_strain.py
```diff
@@ -0,0 +1,159 @@
...
+B1 = B
+B2 = B
+
+# amplitude, angular frequency and wave vector of the strain
+E = 105e-6
+w = 1.3e9 * 2*np.pi
+k = 2*np.pi / 2.8e-6
+
+# simulation grid parameters
+nx, ny, nz = 2000, 250, 1
```
ilateur, line 32: no powers of 2 :smiling_face_with_tear:

examples/rigid_strain.py
ilateur, line 1: `rigid_strain.py` is not the best filename I think... it should at least include "domain wall" (or "DW") and maybe "SAW"? I can't think of the best name right now :thinking:

examples/rigid_strain.py
```diff
@@ -0,0 +1,159 @@
...
+def DW_position(magnet):
+    m_av = magnet.magnetization.average()[0]
+    return m_av*nx*cx / 2
+
+# run the simulation and save the DW postion
+DW_pos = np.zeros(shape=(steps))
+
+for i in range(steps):
+    world.timesolver.run(dt)
+    DW_pos[i] = DW_position(magnet)
```
ilateur, line 148: It takes a few minutes to run, so `tkdm` would be nice

examples/rigid_strain.py
```diff
@@ -0,0 +1,159 @@
...
+# Create the first time term f(x,y,z) = h(x,y,z)*g(t)
+magnet.rigid_shear_strain.add_time_term(shear_strain_t_cos, mask=shear_strain_xyz_cos)
+# Add a second time term to obtain sin and cos
+magnet.rigid_shear_strain.add_time_term(shear_strain_t_sin, mask=shear_strain_xyz_sin)
+
+# Create a domain wall
+magnet.magnetization = twodomain((1,0,0), (0,-1,0), (-1,0,0), nx*cx/2, 10*cx)
+
+# plot the initial and final magnetization
+fig, ax = plt.subplots(nrows=2, ncols=1)
```
ilateur, line 127: `sharex="all"`? maybe `sharey` too?

examples/rigid_strain.py
```diff
@@ -0,0 +1,159 @@
...
+ax1, ax2 = ax
+im_extent = (-0.5*cx*1e6, (nx*cx - 0.5*cx)*1e6, -0.5*cy*1e6, (ny*cy - 0.5*cy)*1e6)
```
ilateur, line 129: maybe this should be a magnet property or something :thinking: we keep copy pasting it everywhere

examples/rigid_strain.py
```diff
@@ -0,0 +1,159 @@
...
+
+# initial magnetization
+ax1.imshow(np.transpose(magnet.magnetization.get_rgb()[:,0,:,:], axes=(1,2,0)), origin="lower", extent=im_extent, aspect="equal")
+ax1.set_title("Initial magnetization")
+ax2.set_title("Final magnetization")
+ax2.set_xlabel("x (µm)")
+ax2.set_ylabel("y (µm)")
```
ilateur, line 136: ax1 should have a ylabel too

examples/rigid_strain.py
```diff
@@ -0,0 +1,159 @@
...
+# final magnetization
+ax2.imshow(np.transpose(magnet.magnetization.get_rgb()[:,0,:,:], axes=(1,2,0)), origin="lower", extent=im_extent, aspect="equal")
+
+plt.show()
+
+# plot DW position in function of time
+plt.plot(np.linspace(0,run,steps)*1e9, (DW_pos)*1e6)
+plt.xlabel("time (ns)")
+plt.ylabel("Domain wall position (µm)")
+plt.show()
```
ilateur, line 159: Should this plot have a title?

ilateur (changes requested):

examples/rigid_strain.py
ilateur: Or just use lambda functions everywhere :thinking:

examples/rigid_strain.py
ilateur: What about `domain_wall_rigid_strain_SAW.py`? too long?

examples/rigid_strain.py
ilateur: It takes a long time :(

examples/rigid_strain.py
DiegoDeGusem: That's a bit long. Maybe shorten `domain_wall` with `dw`/`DW`?

ilateur: Is that clear enough?

examples/rigid_strain.py
DiegoDeGusem: Yeah, I followed the paper. We can just change it

examples/rigid_strain.py
DiegoDeGusem: That is true. These were the sizes used in the paper. But we can be better

ilateur: It seems like the "largeness" is necessary, so dividing cellsize by 2 would mean 4 times the number of cells... A smaller magnet does not seem to support the existence of a domain wall :/

examples/rigid_strain.py
ilateur: And the domain wall tends to move to a side if not placed exactly in the middle :/

ilateur (approved):

ilateur:
Perhaps use "squash and merge" for this branch instead of simply merging.

examples/rigid_strain.py
DiegoDeGusem: I think so?

examples/rigid_strain.py
lamoreel: The system should relax before setting the strain (since at `t = 0s` the strain is non-zero). This should not take long...

examples/rigid_strain.py
lamoreel: I don't like the fact that cells which are **too** big are necessary.

examples/DW_rigid_strain_SAW.py
```diff
@@ -0,0 +1,101 @@
...
+
+# setting DMI to stabilize the DW
+magnet.dmi_tensor.set_interfacial_dmi(1e-3)
+
+# Create a DW
+magnet.magnetization = twodomain((0,0,1), (-1,1,0), (0,0,-1), nx*cx/3, 5*cx)
+
+print("relaxing...")
+magnet.relax()  # relax
+
```
lamoreel, line 43: `minimize` performs much better in this case (70 seconds faster on my machine).

examples/DW_rigid_strain_SAW.py
```diff
@@ -0,0 +1,101 @@
...
+# normal stain, given by exx = E [sin(wt)*cos(kx) - cos(wt)*sin(kx)]
+# Create the first time term
+magnet.rigid_norm_strain.add_time_term(lambda t: (np.sin(w*t), 0., 0.),
+                                       lambda x,y,z: (E*np.cos(k*x), 0., 0.))
+# Add the second time term
+magnet.rigid_norm_strain.add_time_term(lambda t: (np.cos(w*t), 0., 0.),
+                                       lambda x,y,z: (-E*np.sin(k*x), 0., 0.))
+
+# plot the initial and final magnetization
+fig, axs = plt.subplots(nrows=2, ncols=1, sharex="all", sharey="all")
```
lamoreel, line 62: `nrows=1` and `ncols=2` looks better for this vertical geometry.

lamoreel (changes requested):

lamoreel:
The mentioned paper uses Voronoi tessellations to create disorder. Is this something this example might use? :thinking: Perhaps this isn't worth the long simulation time? Or maybe the "disorder percentage" can be an input parameter?

DiegoDeGusem:
I was also thinking about that, but I was scared that it might make it a complex example for users?

examples/DW_rigid_strain_SAW.py
DiegoDeGusem: I find it a bit clearer when they are below one another.
image
image

lamoreel (approved):

examples/DW_rigid_strain_SAW.py
```diff
@@ -0,0 +1,101 @@
...
+ax1, ax2 = axs
+im_extent = (-0.5*cx*1e6, (nx*cx - 0.5*cx)*1e6, -0.5*cy*1e6, (ny*cy - 0.5*cy)*1e6)
+
+# initial magnetization
+ax1.imshow(np.transpose(magnet.magnetization.get_rgb()[:,0,:,:], axes=(1,2,0)), origin="lower", extent=im_extent, aspect="equal")
+ax1.set_title("Initial magnetization")
+ax2.set_title("Final magnetization")
+ax2.set_xlabel("x (µm)")
+ax2.set_ylabel("y (µm)")
+ax1.set_ylabel("y (µm)")
```
ilateur, line 72: I would do the other way around, as `ncols=2`. Only ax1 needs ylabel, and both should get xlabel,

examples/DW_rigid_strain_SAW.py
```diff
@@ -0,0 +1,101 @@
...
+
+# final magnetization
+ax2.imshow(np.transpose(magnet.magnetization.get_rgb()[:,0,:,:], axes=(1,2,0)),
+           origin="lower", extent=im_extent, aspect="equal")
+
+plt.show()
+
+# plot DW position in function of time
+plt.plot(time*1e9, DW_pos*1e6)
+plt.xlabel("time (ns)")
```
ilateur, line 98: (capital T?)

examples/DW_rigid_strain_SAW.py
```diff
@@ -0,0 +1,101 @@
...
+magnet.aex = 1e-11
+magnet.alpha = 0.01
+magnet.ku1 = 8e5
+magnet.anisU = (0,0,1)
+
+# setting DMI to stabilize the DW
+magnet.dmi_tensor.set_interfacial_dmi(1e-3)
+
+# Create a DW
+magnet.magnetization = twodomain((0,0,1), (-1,1,0), (0,0,-1), nx*cx/3, 5*cx)
```
ilateur, line 39: Changing initial domain wall magnetization from `(-1,1,0)` to the symmetric value `(-1,0,0)` cuts down minimization time from 12 seconds down to 7.

PR #63, opened 2025-01-23
DMI standardproblem

DiegoDeGusem:
I have created the 1D, 2D and 3D DMI standarproblems from [here](https://iopscience.iop.org/article/10.1088/1367-2630/aaea1c). For each problem I initialize the system and let it minimize. Then I compare the magnetization obtained in mumax⁺ by the one obtained in mumax³.

### DIFF
--- a/test/mumax3/test_mumax3_DMI_standardproblem1D.py
+++ b/test/mumax3/test_mumax3_DMI_standardproblem1D.py
@@ -53,7 +53,7 @@ def simulations(request):
     magnet.magnetization = magnetization
     magnet.minimize()
     
-    # simulation mumax³
+    # mumax³ simulation
     mumax3sim = Mumax3Simulation(
         f"""
             setcellsize{cellsize}
--- a/test/mumax3/test_mumax3_DMI_standardproblem1D.py
+++ b/test/mumax3/test_mumax3_DMI_standardproblem1D.py
@@ -1,0 +1,84 @@
+"""This test is based on the 1D case in
+   https://iopscience.iop.org/article/10.1088/1367-2630/aaea1c"""
+
+import pytest
+import numpy as np
+from mumaxplus import Ferromagnet, Grid, World
+from mumax3 import Mumax3Simulation
+
+RTOL = 1e-5
+def max_relative_error(result, wanted):
+    err = np.linalg.norm(result - wanted, axis=0)
+    relerr = err / np.linalg.norm(wanted, axis=0)
+    return np.max(relerr)
+
+@pytest.fixture(scope="class", params=[True, False])
+def simulations(request):
+    """This simulates a 1D wire with bulk or interfacial DMI in both
+       mumax³ and mumax⁺ with open boundary conditions."""
+    
+    # test the interfacial and bulk DMI
+    interfacial = request.param
+
+    # constants
+    A = 13e-12
+    D = 3e-3
+    Ku = 0.4e6
+    anisU = (0,0,1)
+    Ms = 0.86e6
+
+    cellsize = (1e-9, 1e-9, 1e-9)
+    gridsize = (100, 1, 1)
+    magnetization = (0,0,1)
+
+    world = World(cellsize=cellsize)
+    magnet = Ferromagnet(world, Grid(gridsize))
+    magnet.enable_demag = False
+    magnet.enable_openbc = True
+
+    magnet.msat = Ms
+    magnet.aex = A
+    magnet.ku1 = Ku
+    magnet.anisU = anisU
+
+    if interfacial:
+        DMI = "Dind"
+        D = -D
+        magnet.dmi_tensor.set_interfacial_dmi(D)
+    else:
+        DMI = "Dbulk"
+        magnet.dmi_tensor.set_bulk_dmi(D)
+    # simulation mumax³
+    mumax3sim = Mumax3Simulation(
+        f"""
+            setcellsize{cellsize}
+            setgridsize{gridsize}
+            msat = {Ms}
+            aex = {A}
+            Ku1 = {Ku}
+            anisU = vector{anisU}
+            {DMI} = {D}
+            enabledemag = False
+            SetPBC(0,2,0)
+
+            m = Uniform{tuple(magnetization)}
+            minimize()
+            saveas(m, "m.ovf")
+        """
+    )
+
+    magnet.magnetization = magnetization
+    magnet.minimize()
+
+    return  magnet, mumax3sim
+
+
+@pytest.mark.mumax3
+class TestDMI1D:
+    """Compare the results of the simulations by comparing the magnetizations.
+    """
+
+    def test(self, simulations):
+        magnet, mumax3sim = simulations
+        err = max_relative_error(magnet.magnetization.eval(), mumax3sim.get_field("m"))
+        assert err < RTOL
--- a/test/mumax3/test_mumax3_DMI_standardproblem1D.py
+++ b/test/mumax3/test_mumax3_DMI_standardproblem1D.py
@@ -31,6 +31,7 @@ def simulations(request):
     gridsize = (100, 1, 1)
     magnetization = (0,0,1)
 
+    # mumax⁺ simulation
     world = World(cellsize=cellsize)
     magnet = Ferromagnet(world, Grid(gridsize))
     magnet.enable_demag = False
@@ -48,6 +49,10 @@ def simulations(request):
     else:
         DMI = "Dbulk"
         magnet.dmi_tensor.set_bulk_dmi(D)
+    
+    magnet.magnetization = magnetization
+    magnet.minimize()
+    
     # simulation mumax³
     mumax3sim = Mumax3Simulation(
         f"""
@@ -66,9 +71,6 @@ def simulations(request):
             saveas(m, "m.ovf")
         """
     )
-
-    magnet.magnetization = magnetization
-    magnet.minimize()
 
     return  magnet, mumax3sim
 
--- a/test/mumax3/test_mumax3_DMI_standardproblem1D.py
+++ b/test/mumax3/test_mumax3_DMI_standardproblem1D.py
@@ -1,5 +1,8 @@
 """This test is based on the 1D case in
-   https://iopscience.iop.org/article/10.1088/1367-2630/aaea1c"""
+   https://iopscience.iop.org/article/10.1088/1367-2630/aaea1c
+   The magnet lies along the z-direction, because mumax³ is not correct. They ignore
+   the z-component.
+   """
 
 import pytest
 import numpy as np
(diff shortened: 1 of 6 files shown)


test/mumax3/test_mumax3_DMI_standardproblem3D.py
```diff
@@ -0,0 +1,107 @@
...
+    magnet = Ferromagnet(world, Grid(gridsize), geometry=geo)
+
+    magnet.enable_demag = False
+    magnet.msat = Ms
+    magnet.aex = A
+    magnet.dmi_tensor.set_bulk_dmi(D)
+
+    magnet.bias_magnetic_field = (0,0,Bz)
+
+    magnet.magnetization = blochskyrmion(magnet.center, 91.5e-9, charge, pol)
```
ilateur, line 48: changing the radius from 91.5e-9 to about 18e-9 makes it so the test "only" takes 14 seconds for me

test/mumax3/test_mumax3_DMI_standardproblem3D.py
```diff
@@ -0,0 +1,107 @@
...
+
+            dx := lx / nx
+            dy := ly / ny
+            dz := lz / nz
+
+            SetGridSize(nx, ny, nz)
+            SetCellSize(dx, dy, dz)
+
+            // Define the cylinder
+            SetGeom(Circle(lx))
```
ilateur, line 72:
```python
f"""
    SetGridSize{gridsize}
    SetCellSize{cellsize}
    SetGeom(Circle({diam}))
"""
```

test/mumax3/test_mumax3_DMI_standardproblem3D.py
```diff
@@ -0,0 +1,107 @@
...
+
+            // External field in T
+            B_ext = vector(0, 0, {Bz})
+
+            // No Demag
+            NoDemagSpins = 1
+
+            m = BlochSkyrmion(1, -1)
+
+            OutputFormat = OVF2_TEXT
```
ilateur, line 86: not needed?

test/mumax3/test_mumax3_DMI_standardproblem2D.py
```diff
@@ -0,0 +1,104 @@
...
+            Ku1         = {Ku}
+            anisU       = vector{tuple(anisU)}
+            Dind        = {D}
+            // No Demag:
+            NoDemagSpins = 1
+
+            // Initial state
+            m = Neelskyrmion({charge}, {pol})
+
+            OutputFormat = OVF2_TEXT
```
ilateur, line 86: not needed?

test/mumax3/test_mumax3_DMI_standardproblem3D.py
ilateur, line 1: We can take the opportunity again to test both open and closed BC.

ilateur (changes requested):
nice that these exist :)

test/mumax3/test_mumax3_DMI_standardproblem1D.py
```diff
@@ -0,0 +1,86 @@
...
+        f"""
+            setcellsize{cellsize}
+            setgridsize{gridsize}
+            msat = {Ms}
+            aex = {A}
+            Ku1 = {Ku}
+            anisU = vector{anisU}
+            {DMI} = {D}
+            enabledemag = False
+            SetPBC(0,2,0)
```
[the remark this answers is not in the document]
ilateur: Comparison of 3 and + for Dbulk with closed BC.
image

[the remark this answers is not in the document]
ilateur: Ah it looks like mumax+ only has open BC. @lamoreel can you confirm or deny?

test/mumax3/test_mumax3_DMI_standardproblem2D.py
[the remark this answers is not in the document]
ilateur: or maybe not

test/mumax3/test_mumax3_DMI_standardproblem3D.py
ilateur: or maybe not

test/mumax3/test_mumax3_DMI_standardproblem1D.py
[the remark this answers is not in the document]
lamoreel: As of now we provide (for magnetodynamic simulations) open, periodic and Neumann boundary conditions. I have no idea what you mean by "closed" :).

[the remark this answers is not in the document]
lamoreel: Neumann boundaries are implemented for every type of exchange interaction (including DMI) and these _should_ be correct... (but this is what tests are for :)).

test/mumax3/test_mumax3_DMI_standardproblem1D.py
```diff
@@ -0,0 +1,86 @@
...
+    magnet.enable_openbc = True
+
+    magnet.msat = Ms
+    magnet.aex = A
+    magnet.ku1 = Ku
+    magnet.anisU = anisU
+
+    if interfacial:
+        DMI = "Dind"
+        D = -D
```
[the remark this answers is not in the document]
DiegoDeGusem: I thought so too. But when I checked the paper and their code they also did this in mumax³ while the other simulators used +D

test/mumax3/test_mumax3_DMI_standardproblem2D.py
```diff
@@ -0,0 +1,104 @@
...
+
+@pytest.mark.mumax3
+class TestDMI2D:
+    """Compare the results of the simulations by comparing the magnetizations.
+    """
+
+    def test(self):
+        magnet, mumax3sim = simulations()
+        err = max_relative_error(magnet.magnetization.eval(), mumax3sim.get_field("m"))
+        assert err < ATOL
```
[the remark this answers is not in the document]
DiegoDeGusem: If we also want to test multiple BC I would keep as a class with then 2 functions. It could also be a `pytest.fixture` here, but then I don't know if that is still possible in the 1D case because the function already has a `fixture` :thinking:. There are also a lot of other tests that use this build

test/mumax3/test_mumax3_DMI_standardproblem1D.py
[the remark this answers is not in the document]
DiegoDeGusem: It works :smile: . The issue lies with mumax³. If the magnet does not lie along the z-direction they ignore that component in the DMI (which seems wrong). I have "fixed" this by rotating the magnet so it lies along the z-direction.

DiegoDeGusem:
I have added 3 more tests for AFM. They still use the same setup as the 3 standard problems. The AFM exchange parameters are set to 0 and both sublattices are initialized in the same way. The magnetization of the AFM is then compared to the FM result.

lamoreel:
I'm not sure how useful these AFM versions are as tests. They don't simulate physical systems and only test part of the used kernel functions. For us it's good to know that they work, but I think we need more formal/complete tests in this case.

ilateur:
They don't test enough, but if it's good for us to know, then it's good that they exist :)

lamoreel:
Yes. Perhaps these can be removed later on when we develop more general ones.

test/mumax3/test_mumax3_DMI_standardproblem1D.py
```diff
@@ -1,5 +1,8 @@
 """This test is based on the 1D case in
-   https://iopscience.iop.org/article/10.1088/1367-2630/aaea1c"""
+   https://iopscience.iop.org/article/10.1088/1367-2630/aaea1c
+   The magnet lies along the z-direction, because mumax³ is not correct. They ignore
```
lamoreel, line 3: "mumax3 is not correct" is a very bold statement. I would be more specific and refer to the issue number in its repo instead.

test/test_DMI_1D_afm.py
```diff
@@ -0,0 +1,109 @@
+"""This test is based on the 3D case in
```
lamoreel, line 1: 1D, not 3D.

test/test_DMI_1D_afm.py
```diff
@@ -0,0 +1,109 @@
...
+    magnet_AFM.afmex_nn = 0
+    magnet_AFM.ku1 = Ku
+    magnet_AFM.anisU = anisU
+
+    if interfacial:
+        for sub in magnet_AFM.sublattices:
+            sub.dmi_tensor.set_interfacial_dmi(D)
+    else:
+        for sub in magnet_AFM.sublattices:
+            sub.dmi_tensor.set_bulk_dmi(D)
```
lamoreel, line 68: Please move the `for` loop one level higher because `for` loops scare me so I only want one :).

test/test_DMI_2D_afm.py
```diff
@@ -0,0 +1,101 @@
+"""This test is based on the 3D case in
```
lamoreel, line 1: 1D.

test/test_DMI_2D_afm.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+ATOL = 1e-4
+def max_absolute_error(result, wanted):
+    err = np.linalg.norm(result - wanted, axis=0)
+    return np.max(err)
+
+def simulations(openBC):
+    """This simulates a 2D circle with interfacial DMI and a Neél skyrmion
+       a ferromagnet and an antiferromagnet."""
```
lamoreel, line 20: This sentence is missing a/some word(s).

test/test_DMI_2D_afm.py
```diff
@@ -0,0 +1,101 @@
...
+    magnet_AFM.aex = A
+    magnet_AFM.afmex_cell = 0
+    magnet_AFM.afmex_nn = 0
+    magnet_AFM.ku1 = Ku
+    magnet_AFM.anisU = anisU
+
+    for sub in magnet_AFM.sublattices:
+        sub.dmi_tensor.set_interfacial_dmi(D)
+    magnet_AFM.sub1.magnetization = neelskyrmion(magnet.center, skyrmion_radius, charge, pol)
+    magnet_AFM.sub2.magnetization = neelskyrmion(magnet.center, skyrmion_radius, charge, pol)
```
lamoreel, line 78: Can't this be set in one line? (`magnet_AFM.magnetization = ...`)

test/test_DMI_3D_afm.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+ATOL = 1e-4
+def max_absolute_error(result, wanted):
+    err = np.linalg.norm(result - wanted, axis=0)
+    return np.max(err)
+
+def simulations(openBC):
+    """This simulates a 3D cylinder with bulk DMI and a bloch skyrmion
+       a ferromagnet and an antiferromagnet."""
```
lamoreel, line 20: Same comment.

test/test_DMI_3D_afm.py
```diff
@@ -0,0 +1,102 @@
...
+    magnet_AFM.aex = A
+    magnet_AFM.afmex_cell = 0
+    magnet_AFM.afmex_nn = 0
+
+    for sub in magnet_AFM.sublattices:
+        sub.dmi_tensor.set_bulk_dmi(D)
+
+    magnet_AFM.bias_magnetic_field = (0,0,Bz)
+
+    magnet_AFM.sub1.magnetization = blochskyrmion(magnet.center, skyrmion_radius, charge, pol)
```
lamoreel, line 75: Same comment.

lamoreel (changes requested):
We should remember that this strategy can also be used to make similar tests with arbitrary DMI tensor elements (so no special cases such as `bulk_dmi`).

lamoreel (approved):

ilateur (approved):

PR #65, opened 2025-02-20
Voronoi enhancement

lamoreel:
Included changes (so far):
* Voronoi tessellator seed is now user accessible and settable.
* `regionOf(coo)` is now wrapped so one can generate indices on the fly
    * Useful for a future moving window.
    * Useful for a future compatibility with PBC.
* Generated region indices are now seed dependent and thus consistent.

### DIFF
--- a/mumaxplus/util/voronoi.py
+++ b/mumaxplus/util/voronoi.py
@@ -3,7 +3,7 @@ import numpy as _np
 import _mumaxpluscpp as _cpp
 class VoronoiTessellator:
 
-    def __init__(self, grainsize, max_idx=256, seed=1234567):
+    def __init__(self, grainsize, seed=None, max_idx=256):
         """Create a Voronoi tessellator instance.
         
         This class is used to generate a Voronoi tessellation, which can
@@ -17,15 +17,16 @@ class VoronoiTessellator:
         ----------
         grainsize : float
             The average grain diameter.
+        seed : int (default=None)
+            The seed of the used random number generators. This seed affects
+            the values of the generated region indices and the number and
+            positions of the Voronoi centers.
         max_idx : int (default=256)
             The maximum region index within the tessellation. This value
             has no upper bound.
-        seed : int (default=1234567)
-            The seed of the used random number generators. This seed affects
-            the values of the generated region indices and the number and
-            positions of the Voronoi centers
         """
-        self._impl = _cpp.VoronoiTessellator(grainsize, max_idx, seed)
+        self.seed = seed if seed else _np.random.randint(1234567)
+        self._impl = _cpp.VoronoiTessellator(grainsize, self.seed, max_idx)
 
     def generate(self, world, grid):
         """Generates a Voronoi tessellation.
--- a/mumaxplus/util/voronoi.py
+++ b/mumaxplus/util/voronoi.py
@@ -3,8 +3,8 @@ import numpy as _np
 import _mumaxpluscpp as _cpp
 class VoronoiTessellator:
 
-    def __init__(self, world, grid, grainsize):
-        self._impl = _cpp.VoronoiTessellator(grid._impl, grainsize, world.cellsize)
+    def __init__(self, world, grid, grainsize, seed=1234567):
+        self._impl = _cpp.VoronoiTessellator(grid._impl, grainsize, world.cellsize, seed)
         
     @property
     def tessellation(self):
--- a/mumaxplus/util/voronoi.py
+++ b/mumaxplus/util/voronoi.py
@@ -4,7 +4,7 @@ import warnings as _w
 import _mumaxpluscpp as _cpp
 class VoronoiTessellator:
 
-    def __init__(self, grainsize, seed=None, max_idx=256):
+    def __init__(self, grainsize, seed=None, max_idx=256, region_of_center=None):
         """Create a Voronoi tessellator instance.
         
         This class is used to generate a Voronoi tessellation, which can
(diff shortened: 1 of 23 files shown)


src/core/voronoi.cpp
```diff
@@ -70,8 +69,7 @@ std::vector<Center> VoronoiTessellator::centersInTile(int3 pos) {
         real cx = (pos.x + distReal_(engine_)) * tilesize_;
         real cy = (pos.y + distReal_(engine_)) * tilesize_;
 
-        centers[n] = Center(real3{cx, cy, 0}, centerIdx_);
-        centerIdx_ += 1;
+        centers[n] = Center(real3{cx, cy, 0}, distInt_(engine_));
```
ilateur, line 72: It would be cool if this `distInt_(engine_)` is exposed to the user. It could be some `int regionOfCenter(real3 centerCoo)` function, which by default returns a random region index, but can be customizable, letting you draw "Voronoied"-shapes with the regions.

src/bindings/wrap_voronoi.cpp
```diff
@@ -11,18 +11,22 @@
...
+        .def(py::init<real, unsigned int, int>(),
+                py::arg("grainsize"),
+                py::arg("max_idx"),
+                py::arg("seed"))
+        .def("coo_to_idx", &VoronoiTessellator::regionOf)
         // TODO: create template function (wrap_system.cpp)
-        .def_property_readonly("tessellation", [](VoronoiTessellator t) {
-            unsigned int* tess = t.generate().getHostCopy();
+        .def("generate", [](VoronoiTessellator& t, Grid grid, real3 cellsize) {
+            unsigned int* tess = t.generate(grid, cellsize).getHostCopy();
```
ilateur, line 21: Huh this copying from CPU to GPU to CPU to GPU again is a bit silly :thinking:

mumaxplus/util/voronoi.py
```diff
@@ -3,18 +3,44 @@
...
+        `generate` has been called. E.g. retrieving a list of region
+        indices requires a specified world and grid.
+
+        Parameters
+        ----------
+        grainsize : float
+            The average grain diameter.
+        max_idx : int (default=256)
+            The maximum region index within the tessellation. This value
+            has no upper bound.
```
ilateur, line 22: The *exclusive* maximum?

ilateur (commented):
:)

src/core/voronoi.hpp
```diff
@@ -29,16 +29,16 @@ struct Int3Hash {
 
 class VoronoiTessellator {
  public:
-  VoronoiTessellator(Grid grid, real grainsize, real3 cellsize);
+  VoronoiTessellator(real grainsize, unsigned int maxIdx=256, int seed=1234567);
```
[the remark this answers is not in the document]
lamoreel: If we want simulations to be consistent/reproducible, then a fixed default value is needed (of course this value can be random).

src/bindings/wrap_voronoi.cpp
lamoreel: No? The tessellation array is generated and saved on the GPU and gets copied here to the CPU once. Than this buffer is cast directly into a Python array.

ilateur: If I got it right, the C++ `generate` function fills a CPU C++ array, then copies to the GPU to return a GpuBuffer. This binding then gets the hostcopy to translate back to CPU C++ array, which is then converted into a numpy array to be used in Python by the user. The user then gives it to the relevant Magnet, where the `add_((anti)ferro)magnet` function casts the python array back to a GpuBuffer.

When Python is the middleman, the first GpuBuffer conversion is a bit silly.

src/bindings/wrap_voronoi.cpp
ilateur, line 1: It would be superdupercool if `centerIdx`/`region_of_center` is a (settable) property (a property and also a function may be weird :thinking:) or would at least be accessible

Or is that... not useful at all?
would be cool tho

src/bindings/wrap_voronoi.cpp
```diff
@@ -11,28 +11,32 @@
...
-            return py::array_t<unsigned int>(shape, strides, tess, free_when_done);
-            });
+            ssize_t shape[3] = {size.z, size.y, size.x};
+            ssize_t strides[3] = {
+                static_cast<ssize_t>(sizeof(unsigned int)) * size.x * size.y,
+                static_cast<ssize_t>(sizeof(unsigned int)) * size.x,
+                static_cast<ssize_t>(sizeof(unsigned int))
+            };
+            return py::array_t<unsigned int>(shape, strides, raw, free_when_done);
+        });
```
ilateur, line 41: :O you actually did it

Now what hoops would a mumaxC++ user have to jump through to convert the output of `generate` to a usable `GpuBuffer` for the regions of a `Magnet`?

src/core/voronoi.cpp
```diff
@@ -5,17 +5,26 @@
...
+                                       unsigned int maxIdx,
+                                       const std::function<unsigned int(real3)>& centerIdx)
     : grainsize_(grainsize),
       seed_(seed),
       distReal_(0.0, 1.0),
       distInt_(0, maxIdx) {
+        if (centerIdx) { centerIdx_ = centerIdx; }
+        else {
+            centerIdx_ = [this](real3 coo) -> unsigned int {
+                return distInt_(engine_);
```
ilateur, line 19: Could the `real3 coo` somehow be used as a seed for the random number? Or are C++ random number generators too annoying?

src/bindings/wrap_voronoi.cpp
ilateur: Oh I see, literally just converting raw output data to `GpuBuffer<unsigned int>`, nice!

ilateur (changes requested):
Two small meaningful comments, the rest are just thoughts :)
but I like

src/core/voronoi.cpp
lamoreel: This is possible, but I'm not sure why this would be more optimal... As of now, the tile positions are used as a seed. It might be overkill to reset this seed for every individual `coo`.

DiegoDeGusem (approved):

PR #66, opened 2025-02-25
Voronoi PBC

lamoreel:
Implementation of periodic boundary conditions in the Voronoi tesselator.

Periodicity implies that tiles can no longer (partially) exist outside `Grid`. This leads to a fudging of the tile size, so all tiles fit evenly, and the generalization to rectangular tiles.
This might not be the best option and there are surely edge cases where this logic breaks, but then again, the use of Voronoi regions is only physical with _decent_ parameters (grid size and grain size).

### DIFF
--- a/mumaxplus/util/voronoi.py
+++ b/mumaxplus/util/voronoi.py
@@ -37,7 +37,7 @@ class VoronoiTessellator:
 
         # is this the cleanest way to check PBC?
         # do we need this check in the C++ module?
-        has_pbc = world.mastergrid != Grid((0,0,0)) and world.pbc_repetitions != (0,0,0)
+        has_pbc = world.pbc_repetitions != (0,0,0)
 
         self.tessellation = self._impl.generate(grid._impl, world.cellsize, has_pbc)
 
@@ -45,7 +45,13 @@ class VoronoiTessellator:
     
     def coo_to_idx(self, x, y, z):
         """Returns the region index (int) of the given coordinate within the
-        Voronoi tessellation."""
+        Voronoi tessellation.
+
+        **Important:** This method has no information about the used world and
+        grid. E.g. this means that periodic boundary conditions will not apply.
+        This can be overriden by calling `generate` before assinging this function
+        to the `Magnet`'s regions parameter.
+        """
         return self._impl.coo_to_idx((x,y,z))
 
     @property
--- a/mumaxplus/util/voronoi.py
+++ b/mumaxplus/util/voronoi.py
@@ -1,6 +1,8 @@
 import numpy as _np
 
 import _mumaxpluscpp as _cpp
+from mumaxplus.world import World
+from mumaxplus.grid import Grid
 class VoronoiTessellator:
 
     def __init__(self, grainsize, max_idx=256, seed=1234567):
@@ -32,7 +34,12 @@ class VoronoiTessellator:
 
         Returns an ndarray of shape (nz, ny, nx) which is filled
         with region indices."""
-        self.tessellation = self._impl.generate(grid._impl, world.cellsize)
+
+        # is this the cleanest way to check PBC?
+        # do we need this check in the C++ module?
+        has_pbc = world.mastergrid != Grid((0,0,0)) and world.pbc_repetitions != (0,0,0)
+
+        self.tessellation = self._impl.generate(grid._impl, world.cellsize, has_pbc)
 
         return self.tessellation
     
--- a/src/core/voronoi.cpp
+++ b/src/core/voronoi.cpp
@@ -15,9 +15,31 @@ VoronoiTessellator::VoronoiTessellator(real grainsize, unsigned int maxIdx, int 
         lambda_ = tilesize_in_grains * tilesize_in_grains;
     }
 
+    real VoronoiTessellator::findTileSize(real grid_size, real grainsize) {
+        real min_tile_size = 2 * grainsize;
+        unsigned int N = std::max(1u, static_cast<unsigned int>(std::ceil(grid_size / min_tile_size)));
+        return grid_size / N;  // Adjusted tile size to fit the grid evenly
+    }
+
+
 GpuBuffer<unsigned int> VoronoiTessellator::generate(Grid grid, real3 cellsize) {
 
+    gridsize_ = grid.size();
+    cellsize_ = cellsize;
+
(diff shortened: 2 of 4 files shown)


src/core/voronoi.cpp
```diff
@@ -49,6 +79,13 @@ unsigned int VoronoiTessellator::regionOf(real3 coo) {
     return nearest.ridx;
 }
 
+real3 VoronoiTessellator::periodicShift(real3 coo, real3 center) {
+    real3 d = center - coo;
+    d.x -= grid_dims_.x * (std::abs(d.x) > 0.5 * grid_dims_.x ? std::copysign(1.0, d.x) : 0);
+    d.y -= grid_dims_.y * (std::abs(d.y) > 0.5 * grid_dims_.y ? std::copysign(1.0, d.y) : 0);
+    return coo + d;
```
ilateur, line 86: This seems a bit better?
```cpp
real3 d = center - coo;
if (std::abs(d.x) > 0.5 * grid_dims_.x) center.x -= std::copysign(grid_dims_.x, d.x);
if (std::abs(d.y) > 0.5 * grid_dims_.y) center.y -= std::copysign(grid_dims_.y, d.y);
return center;
```
Three changes:
- no multiplication, only use copysign
- no -=0
- no extra addition, just directly modify and return center (it's a local copy anyway)

src/core/voronoi.hpp
```diff
@@ -52,7 +56,13 @@ class VoronoiTessellator {
   GpuBuffer<unsigned int> tessellation;
 private:
   real grainsize_;
-  real tilesize_;
+  real3 grid_dims_;
+  bool pbc_;
+
+  real3 tilesize_;
+  int numTiles_x;
+  int numTiles_y;
```
ilateur, line 64: I don't really like adding class variables which are only given a value in some cases (like with PBC here), but I get it

src/core/voronoi.hpp
```diff
@@ -33,11 +33,15 @@ class VoronoiTessellator {
   ~VoronoiTessellator() = default;
 
   // * Generate a Voronoi tessellation
-  GpuBuffer<unsigned int> generate(Grid grid, real3 cellsize);
+  GpuBuffer<unsigned int> generate(Grid grid, real3 cellsize, const bool pbc);
```
ilateur, line 36: It would be nice if pbc is a "`bool3`", so it is toggleable per direction :grin:

mumaxplus/util/voronoi.py
```diff
@@ -32,13 +34,24 @@ def generate(self, world, grid):
 
         Returns an ndarray of shape (nz, ny, nx) which is filled
         with region indices."""
-        self.tessellation = self._impl.generate(grid._impl, world.cellsize)
+
+        # is this the cleanest way to check PBC?
```
ilateur, line 38: I guess checking the `mastergrid` would be slightly better? But there is no simple True or False answer to this question

mumaxplus/util/voronoi.py
```diff
@@ -32,13 +34,24 @@ def generate(self, world, grid):
...
+        # do we need this check in the C++ module?
```
ilateur, line 39: I don't think so? I would leave it up to the user to enable Voronoi PBC, independent of the used world

ilateur (changes requested):
Nice changes, but with some concerns, which might be nothing :)

src/core/voronoi.hpp
lamoreel: Me neither, but the alternative is calculating these values multiple times if PBC are enabled (using `std::ceil` in each direction), so storing 2 integers seems more efficient.

src/core/voronoi.cpp
```diff
@@ -10,14 +11,35 @@ VoronoiTessellator::VoronoiTessellator(real grainsize, unsigned int maxIdx, int
...
+        pbc_ = false;
+      }
+
+real3 VoronoiTessellator::getTileSize(real3 griddims) {
+
+    if (pbc_)
+        return real3{2 * grainsize_, 2 * grainsize_ , 0};
+
+    return griddims / real3{std::max(1.0f, std::ceil(griddims.x / (2 * grainsize_))),
+                            std::max(1.0f, std::ceil(griddims.y / (2 * grainsize_))),
```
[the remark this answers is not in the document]
lamoreel: Indeed, as I understand it
- `lambda_` will be reduced if one dimension is shorter than the other (and vice versa, of course). So there will be less centers in smaller tiles leading to relatively larger grains (compared to an unadjusted value for `lambda_`).
- The center coordinates remain uniformly distributed, but the grains themselves aren't confined to a certain tile in which its center happens to lie, so the shape of the grains is unaffected by this.

src/core/voronoi.hpp
lamoreel: I"ll keep this in mind when I extend the functionalities to 3D :).

src/core/voronoi.cpp
```diff
@@ -35,10 +57,18 @@ unsigned int VoronoiTessellator::regionOf(real3 coo) {
     real mindist = INFINITY;
     for (int tx = t.pos.x-1; tx <= t.pos.x+1; tx++) {
         for (int ty = t.pos.y-1; ty <= t.pos.y+1; ty++) {
-            std::vector<Center> centers = centersInTile(int3{tx, ty, 0});
+
+            int wrapped_tx = pbc_ ? (tx + numTiles_x) % numTiles_x : tx;
+            int wrapped_ty = pbc_ ? (ty + numTiles_y) % numTiles_y : ty;
```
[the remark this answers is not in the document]
lamoreel: Now that you mention it, the function `tileOfCell(coo)` also assumes an unaltered origin. This means that the whole Voronoi machinery fails when a non-default origin is set, even for non-pbc simulations. I suggest we put a pin in this and take this up in another branch?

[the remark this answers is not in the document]
lamoreel: After some good looking at the code, my comment above is **not true**, entirely. `generate` assumes no properties of the used grid, apart from its size (in each direction). So the tessellation is build up independently from the _true_ coordinates of cells. This should pose no problem when the window is moving (because the tessellator only cares about the size of the window).
Then again, `regionOf(real3 coo)` **does** place the origin of the grid at `(0,0,0)`, so this function would become useless when we have a moving window.
Nevertheless, I suppose these issues are to be treated in another branch, since this problem extends beyond the functionalities that this one wants to add.

PR #67, opened 2025-03-11
Fix cmd

lamoreel:
This PR fixes the `cmd` executable (allows for easy C++/CUDA debugging). The executables are installed inside the activated conda environment when using `pip` to compile `mumaxplus`.

It might be possible that a clean cache/build is necessary. In order to do this, run
```
pip uninstall -y mumaxplus
rm -rf build dist *.egg-info
pip install .
```
Then, the executable can be run by executing
```
cmd
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -66,11 +66,8 @@ class CMakeBuild(build_ext):
             ["cmake", ext.sourcedir] + cmake_args, cwd=self.build_temp, env=env
         )
         subprocess.check_call(
-            ["cmake", "--build", ".", "--target", "_mumaxpluscpp"] + build_args,
+            ["cmake", "--build", ".", "--target", "_mumaxpluscpp", "install"] + build_args,
             cwd=self.build_temp,
-        )
-        subprocess.check_call(
-            ["cmake", "--build", ".", "--target", "install"] + build_args, cwd=self.build_temp
         )
 
 setup(
--- a/setup.py
+++ b/setup.py
@@ -69,7 +69,9 @@ class CMakeBuild(build_ext):
             ["cmake", "--build", ".", "--target", "_mumaxpluscpp"] + build_args,
             cwd=self.build_temp,
         )
-
+        subprocess.check_call(
+            ["cmake", "--build", ".", "--target", "install"] + build_args, cwd=self.build_temp
+        )
 
 setup(
     name="mumaxplus",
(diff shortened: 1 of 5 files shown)


src/cmd/standard_problem4.cpp
```diff
@@ -22,7 +22,9 @@ void standard_problem4() {
 
   MumaxWorld mWorld(cellsize);
   Grid mGrid(n);
-  auto magnet = mWorld.addFerromagnet(mGrid, 3, ferromagnet_name);
+  GpuBuffer<bool> geometry;
+  GpuBuffer<unsigned int> regions;
+  auto magnet = mWorld.addFerromagnet(mGrid, geometry, regions, ferromagnet_name);
```
ilateur, line 27: Huh... inconvenient :thinking:

src/cmd/CMakeLists.txt
```diff
@@ -9,3 +9,5 @@ target_include_directories(cmd PRIVATE ${CMAKE_CUDA_TOOLKIT_INCLUDE_DIRECTORIES}
 
 target_link_libraries(cmd PUBLIC physics)
 target_link_libraries(cmd PUBLIC core)
+
+install(TARGETS cmd DESTINATION "$ENV{CONDA_PREFIX}/bin")
```
ilateur, line 13: This feels like a very random destination... At first I was looking for a `cmd` executable file somewhere in `mumaxplus/` or `mumaxplus/build/`, but it just gets installed as a command in the conda environment..?

setup.py
```diff
@@ -69,7 +69,9 @@ def build_extension(self, ext):
             ["cmake", "--build", ".", "--target", "_mumaxpluscpp"] + build_args,
             cwd=self.build_temp,
         )
-
+        subprocess.check_call(
+            ["cmake", "--build", ".", "--target", "install"] + build_args, cwd=self.build_temp
+        )
```
ilateur, line 74: These two commands look incredibly similar :monocle_face:

ilateur (commented):
I wonder how Jeroen and Oleh used to compile the `cmd` executable... some manual CMake magic?

src/cmd/CMakeLists.txt
lamoreel: Installation in the `bin` of the `mumaxplux` `conda` environment lets users run the `cmd` command from anywhere as long as the `env` is activated. Where would you like it to be installed instead?

ilateur: I would have thought inside the `mumaxplus` or `mumaxplus/build` directory... but perhaps it is most convenient to call mumaxC++ from anywhere... I don't really know

ilateur (approved):

DiegoDeGusem (approved):

PR #68, opened 2025-04-08
Improve accuracy of traction-free BC to 2nd order

ilateur:
Previously, the traction-free boundary conditions used to be applied using a central difference scheme, combined with an average of two cells as approximation. So for the cell at the left boundary at index `i`, the derivative in the x direction was approximated by a central difference using a step size of only half a cell:
```
dσ/dx ≈ (σ(i + ½) - σ(i - ½)) / (2 c/2)
```
The left side is determined by the boundary conditions: σ(i - ½) = 0
The right side is in between cells. To approximate this value, we use the average of the known cells:
```
σ(i + ½) ≈ (σ(i + 1) + σ(i)) / 2
```
Which yielded the previous derivative at the boundary:
```
dσ/dx ≈ (σ(i + 1) + σ(i)) / (2 c)
```

The new way uses a custom 2nd order accurate stencil:
```
dσ/dx = (σ(i + 1) + 3 σ(i) - 4 σ(i - ½)) / (3 c) + O(c²)
```
Then `σ(i - ½) = 0` is is used to impose the boundary condition.

Besides being more correct than before, this way also seems to be much more stable. Sadly, it is not magic, and it is still possible to have unstable simulations.

### DIFF
--- a/src/physics/elasticforce.cu
+++ b/src/physics/elasticforce.cu
@@ -15,14 +15,16 @@ __device__ int3 tensorRowComps(int row) {
   return int3{tensorComp(row, 0), tensorComp(row, 1), tensorComp(row, 2)};
 }
 
-// Numerical divergence of stress with central five-point stencil in bulk material.
-// Lower order accuracy (three-point stencil) central difference is used in bulk
-// or 1 cell away from boundary.
-// At the boundary, traction-free bondary conditions are implemented, inspired
-// by a three-point stencil with step size h equal to half cellsize.
+/**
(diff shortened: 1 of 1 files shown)


lamoreel (approved):

DiegoDeGusem:
What if you try an even higher order stencil? Would that be magical :thinking:

DiegoDeGusem (approved):

ilateur:
I can try, but I'd have to calculate one first. Perhaps it's better, but I doubt it will be fully stable.

PR #69, opened 2025-04-10
Asymptotic Demag Calculations

DiegoDeGusem:
To fix the catastrophic cancelation issue with the analytical demag calculation I have added the possibility to calculate the asymptotic approximation of the demagnetization tensor. This method is based on the same procedure used in [OOMMF](https://math.nist.gov/~MDonahue/talks/mmm2020-talk.pdf).
image
Based on a given order (default 11) all terms for the asymptotic expansion get determined in `demagasymptotic.cpp`. This returns a big vector containing all values needed for the calculation which are then used in `demagasymptotic.cu`.
The distance at which the code switches from the analytical result to the asymptotic calculation is also based on the method used in OOMMF. There they assumed an error for the analytical result (E_an) and an error for the asymptotic result (E_as), so the switching point is chosen to be where E_an / E_as crosses 1.


src/physics/strayfieldkernel.cu
```diff
@@ -37,13 +44,41 @@ __global__ void k_strayFieldKernel(CuField kernel, const Grid mastergrid,
...
+
+        // Assume the following errors on the analytical and asymptotic result
+        // E_analytic = eps R³/V
+        // E_asymptotic = V R²/(5(R²-dmax²)) dmax^(n-3)/R^(n)
+        // Here V is dx*dy*dz, dmax = max(dx,dy,dz), n is the order of asymptote
+        // and eps is a constant determined by trial and error.
+        // Use the analytical model when
+        // E_analytic / E_asymptotic < 1
+
+        // Todo: make the order variable?
```
lamoreel, line 57: Done :smile: .

src/physics/strayfieldkernel.cu
```diff
@@ -37,13 +44,41 @@ __global__ void k_strayFieldKernel(CuField kernel, const Grid mastergrid,
...
-          int3 coo_ = coo + (int3{i,j,k} * mastergrid.size());
+        int3 coo_ = coo + (int3{i,j,k} * mastergrid.size());
+
+        // Assume the following errors on the analytical and asymptotic result
+        // E_analytic = eps R³/V
+        // E_asymptotic = V R²/(5(R²-dmax²)) dmax^(n-3)/R^(n)
+        // Here V is dx*dy*dz, dmax = max(dx,dy,dz), n is the order of asymptote
+        // and eps is a constant determined by trial and error.
+        // Use the analytical model when
+        // E_analytic / E_asymptotic < 1
```
lamoreel, line 55: Where do these expressions come from? What's the used value for `eps`? Please add some refs.

src/physics/strayfieldkernel.cu
```diff
@@ -37,13 +44,41 @@ __global__ void k_strayFieldKernel(CuField kernel, const Grid mastergrid,
...
+
+        // Todo: make the order variable?
+        // Todo: switch to O(R⁻³) when far away?
+
+        double x = coo_.x * cellsize.x;
+        double y = coo_.y * cellsize.y;
+        double z = coo_.z * cellsize.z;
+        double R = sqrtf(x*x + y*y + z*z);
+        double V = cellsize.x * cellsize.y * cellsize.z;
+        double h = fmax(cellsize.x,fmax(cellsize.y,cellsize.z));
```
lamoreel, line 65: You explicitly use `double`, which raises two concerns;
- Is this wanted when compiled with single precision?
- Why do you use `sqrtf` which casts the result into a `float` (and then again into `double` to assign the value to `R`)? `sqrt` keeps the double precision.

test/test_demag.py
```diff
@@ -43,3 +43,52 @@ def test_demagfield(self):
         result = magnet.demag_field.eval()
         err = np.max(np.abs((wanted - result) / result))
         assert err < 2e-3
+
+    # Compare the demagkernel with high accurate json files
+    def test_Nxx(self):
+        nx, ny, nz = 100, 100, 1
```
lamoreel, line 49: What about 3D simulations? Or is there no new math in these?

src/physics/demagasymptotic.hpp
```diff
@@ -0,0 +1,85 @@
...
+void combinationsRecursive(
+    const std::vector<int>&,
+    int,
+    int,
+    std::vector<int>&,
+    std::vector<std::vector<int>>&
+);
+
+std::vector<std::vector<int>> derivativeCombinations(int, int);
+std::vector<int> uptoOrder(int, std::vector<std::vector<int>>);
```
lamoreel, line 21: These might need some comments explaining what the functions do. Now it's just a hodgepodge of `std::vector` and `int` :smile: .

src/physics/demagasymptotic.cpp
```diff
@@ -0,0 +1,235 @@
+#include "demagasymptotic.hpp"
+
+#include <iostream>
+#include <vector>
+
+/* Calculate all needed derivatives for the asymptotic expansion and put the
+   result in one list.
```
lamoreel, line 7: It's a vector, not a list :).

src/physics/demagasymptotic.cpp
```diff
@@ -0,0 +1,235 @@
...
+            hx = term[5];
+            hy = term[6];
+            hz = term[7] + 1;
+            new_expansion.push_back({N,a,b,c,P,hx,hy,hz});
+        }
+    }
+    return cleanup(new_expansion);
+}
+
+// Clean up by adding N values of terms with the same qoefficients
```
lamoreel, line 141: Typo coefficients.

src/physics/demagasymptotic.cpp
```diff
@@ -0,0 +1,235 @@
...
+        if (total > 0 && total <= max_order)
+            result.push_back(current);
+        return;
+    }
+
+    for (int val : even_orders) {
+        current[depth] = val;
+        combinationsRecursive(even_orders, max_order, depth + 1, current, result);
+    }
+}
```
lamoreel, line 186: These indentations and (lack of) newlines hurt my brain

src/physics/demagasymptotic.cpp
```diff
@@ -0,0 +1,235 @@
...
+    std::vector<std::vector<int>> valid_combinations;
+    std::vector<int> current(num_variables, 0);
+
+    combinationsRecursive(even_orders, max_order, 0, current, valid_combinations);
+
+    return valid_combinations;
+}
+
+// Generate all terms up to a specific order (also including that order).
+std::vector<int> uptoOrder(int order, std::vector<std::vector<int>> expansion) {
```
lamoreel, line 205: `to` should have follow a camel (`upToOrder`).

src/physics/demagasymptotic.cpp
```diff
@@ -0,0 +1,235 @@
...
+    int num_variables = 3;
+    std::vector<std::vector<int>> combos = derivativeCombinations(num_variables, order);
```
lamoreel, line 207: This `num_variables` variable is only used to pass to `derivativeCombinations` so can be omitted (it's very dangerous to hard code this kind of stuff :)). Also, can the need for this argument as a whole be omitted as well?

src/physics/demagasymptotic.cpp
```diff
@@ -0,0 +1,235 @@
...
+        }
+        for (int i = 0; i < der[2]; i++){  // z derivatives
+            new_expansion = dz(new_expansion);
+        }
+
+        for (std::vector<int>& term : new_expansion) {
+            for (const int val : term) {
+                result.push_back(val);
+            }
+        }
```
lamoreel, line 232: This looks like funny code... `new_expansion` is given a new value in each iteration of each `for` loop. What am I missing?

src/physics/demagasymptotic.cu
```diff
@@ -0,0 +1,128 @@
+#include "demagasymptotic.hpp"
+
+#include "gpubuffer.hpp"
+#include <cstdio>
+
+
+// Todo: Allow a user to chose an order?
```
lamoreel, line 7: Done.

src/physics/demagasymptotic.cu
```diff
@@ -0,0 +1,128 @@
...
+
+// Function to calculate factorials
+__host__ __device__ int fac(int val) {
+  double result = 1;
```
lamoreel, line 11: Why do you need `double` if you only work with integers?

src/physics/demagasymptotic.cu
```diff
@@ -0,0 +1,128 @@
...
+ */
+__host__ __device__ real calcAsymptoticNxx(int3 idx, real3 cellsize, int* expansionNxxptr, size_t sizeNxx) {
+
+  double hx = cellsize.x;
+  double hy = cellsize.y;
+  double hz = cellsize.z;
+  double x = idx.x * hx;
+  double y = idx.y * hy;
+  double z = idx.z * hz;
+  double R = sqrtf(x*x + y*y + z*z);
```
lamoreel, line 30: Same comment about `double` and `sqrtf`.

src/physics/demagasymptotic.cu
```diff
@@ -0,0 +1,128 @@
...
+    int e = expansionNxxptr[i+6];
+    int f = expansionNxxptr[i+7];
+
+    double A = fac(d+2) * fac(e+2) * fac(f+2);
+    double weight = N/A;
+
+    double hi = pow(hx,d)*pow(hy,e)*pow(hz,f);
+    double xyz = pow(x,a)*pow(y,b)*pow(z,c);
+
+    double Rpow = pow(R,P);
```
lamoreel, line 50: `double` or `real` (which is equal to `double` when compiled with double precision)?

src/physics/demagasymptotic.cu
```diff
@@ -0,0 +1,128 @@
...
+    result *= i;
+  }
+  return result;
+}
+
+/** Calculate the asymptotic solution of the demagkernel component Nxx. 
+    This method is based on the method used in OOMMF.
+    https://math.nist.gov/~MDonahue/talks/mmm2020-talk.pdf
+ */
+__host__ __device__ real calcAsymptoticNxx(int3 idx, real3 cellsize, int* expansionNxxptr, size_t sizeNxx) {
```
lamoreel, line 22: Is this really a `__device__` function?

src/physics/strayfieldkernel.cu
```diff
@@ -37,13 +44,41 @@ __global__ void k_strayFieldKernel(CuField kernel, const Grid mastergrid,
...
+        
+        if (5 * 1e-10 * (R*R - h*h)/(V*V) * pow(R,order+1)/pow(h,order-3) < 1) {
```
lamoreel, line 67: I wonder if we need the user to be able to set this. Something like "use Newell up to 5 cells, asymptotic otherwise" with "5" being something settable.

src/bindings/wrap_strayfield.cpp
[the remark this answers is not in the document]
DiegoDeGusem: Yes, this might be a problem. When you are trying to do a pure demag calculation, you can set the order and method using
```
from mumaxplus import StrayField

field = StrayField(magnet, grid)
field.set_method("fft")
field.set_order(21)
```
I am not sure if it is even possible to change it for a "normal" simulation :thinking:

src/physics/strayfieldkernel.cu
DiegoDeGusem: I also thought about this, but the analytical calculations for the demag kernel also uses doubles. To further complicate this, depending on the use of singles or doubles, the value of epsilon should also change. To avoid this, I use doubles consistently for a high accurate demag kernel.

test/test_demag.py
DiegoDeGusem: All the new math is contained within the calculation of the Nxx and Nxy component of the demag kernel, so I don't think 3D is needed. But I can add it if you want me to.

src/physics/demagasymptotic.cpp
DiegoDeGusem: So, you have an entire ~~list~~ vector containing vectors with 3 values. The first value has the amount of x derivatives, the second the amount of y derivatives and the third the amount of z derivatives for each term in the expansion. Your function needs  to go through each one of these, because `(d²/dx² + d²/dy²) f(x,y,z) = d²/dx² f(x,y,z) + d²/dy² f(x,y,z)`. So each time we calculate a new expansion term, we need to start with `f(x,y,z)` and that is why we start with `new_expansion = expansion` at the beginning of each loop.
The functions `dx`, `dy` and `dz` calculate the derivatives of the input and return the result as given in the math screenshot.
Each time you take a derivative you should continue working with that result until all derivatives have been calculated. For example `d²/dx² x²`: `d/dx x² = 2x` --> `d/dx 2x = 2`.
So that's why `new_expansion` is used as an input and then that result is used as `new_expansion`.
Finally we put the result in one giant vector, because gpubuffer cannot work with a vector of vectors.

src/physics/strayfieldkernel.cu
DiegoDeGusem: hmm, doesn't it make more sense that the computer calculates the optimal switching radius?

src/physics/demagasymptotic.cu
DiegoDeGusem: double is also what the newell calcualtions use for the demag kernel.

src/physics/demagasymptotic.cu
DiegoDeGusem: I thought so, because it gets called by the kernel function (just like the Newell calcualtions).

src/physics/strayfieldkernel.cu
lamoreel: Hmm, okay, but what if, for example, someone wants to compare our Newell to their own implementation of Newell, then this comparison is limited to some obscure (unknown) number of cells.
Also, it fits with our plus-mindset that this threshold (if condition) is accessible and settable :).

DiegoDeGusem: Ok true, we can use this as a standard value and let a user choose something if they want to.

src/physics/demagasymptotic.cu
lamoreel: Right, in `strayfieldkernel.cu`, my bad!

src/physics/demagasymptotic.hpp
```diff
@@ -3,22 +3,54 @@
...
 #include "gpubuffer.hpp"
 
+/** 
+ *  The derivatives dx, dy and dz of a function of the form
+ *  f(x,y,z) = N * hx^f * hy^g * hz^i * x^a * y^b * z^c / R^P 
+ *  where N is a constant, R = sqrt(x² + y² + z²), hx, hy and hz are cell sizes
+ *  and x, y and z are coordinates.
+ *  These functions can be rewritten as vectors (N,a,b,c,P,f,g,i).
+ *  dx, dy and dz accept a vector containing vectors of this form return the
+ *  derivativesand as a vector containing vectors of that same form
```
lamoreel, line 13: Extra space between first two words.

src/physics/strayfieldkernel.cu
```diff
@@ -46,25 +46,26 @@ __global__ void k_strayFieldKernel(CuField kernel, const Grid mastergrid,
...
-        // Here V is dx*dy*dz, dmax = max(dx,dy,dz), n is the order of asymptote
-        // and eps is a constant determined by trial and error.
-        // Use the analytical model when
-        // E_analytic / E_asymptotic < 1
+        /* Determine the switching the same way as in the OOMMF code.
+        Assume the following errors on the analytical and asymptotic result
+        E_analytic = eps R³/V
+        E_asymptotic = V R²/(5(R²-dmax²)) dmax^(n-3)/R^(n)
+        Here V is dx*dy*dz, dmax = max(dx,dy,dz), n is the order of asymptote
+        and eps = 5e-10 is a constant determined by trial and error.
```
lamoreel, line 54: I'm wondering if `eps` should be accessible by the user :thinking: .

test/test_demag.py
lamoreel: Maybe, yes, just in case the machinery changes in the future.

lamoreel (changes requested):

mumaxplus/strayfield.py
```diff
@@ -42,3 +42,16 @@ def set_order(self, order):
         """
         assert isinstance(order, int), "The order should be an integer."
         self._impl.set_order(order)
+
+    def set_switch_radious(self, R):
```
lamoreel, line 46: Typo "radius".

mumaxplus/strayfield.py
```diff
@@ -42,3 +42,16 @@ def set_order(self, order):
...
+        """Set the radious from which the asymptotic expantion should be used.
+        Default is -1, then the OOMMF method is used:
```
lamoreel, line 48: Default value should be apparent in line 46.

src/physics/strayfieldfft.hpp
```diff
@@ -37,6 +37,9 @@ class StrayFieldFFTExecutor : public StrayFieldExecutor {
   /** Return the asymptotic computation order. */
   int order() const { return kernel_.order(); }
 
+  /** Return the aswitching radious. */
```
lamoreel, line 40: Rogue "a".

ilateur:
> Not to be a pain, but it would also be very cool if the user could set a `switching_condition`; a boolean function which gets evaluated inside `strayfieldkernel.cu` (for inspiration, maybe compare with `timesolver.run_while()`).

Sadly, this function has to be evaluated per cell on device (a `__device__` function), which has to be defined at compile-time as far as I understand, so this is not dynamically user-settable.

ilateur:
Also I *will* get to this pull request eventually. It is just intimidatingly large.

src/bindings/wrap_strayfield.cpp
```diff
@@ -21,5 +21,9 @@ void wrap_strayfield(py::module& m) {
         } else {
           throw std::invalid_argument("Method should be \"fft\" or \"brute\"");
         }
-      });
+      })
+      .def("set_order", &StrayField::setOrder, py::arg("order"),
+           "Set the order of 1/R in the asymptotic expansion of the demag kernel.")
+      .def("set_switching_radius", &StrayField::setSwitchingradius, py::arg("switching radius"),
+           "Set the radius at which the asymptotic expansion should be used.");
```
lamoreel, line 28: "from which"

test/test_demag.py
```diff
@@ -43,3 +43,59 @@ def test_demagfield(self):
         result = magnet.demag_field.eval()
         err = np.max(np.abs((wanted - result) / result))
         assert err < 2e-3
+
+    # Compare the demagkernel with high accurate json files. These were made
+    # with the BigFloat package with an accuracy of 1024 bits
+    # and the analytical method.
+    def test_Nxx(self):
+        nx, ny, nz = 100, 100, 10
```
lamoreel, line 51: Is the math different when `nx != ny`? Otherwise these tests can be made even more general.

test/test_demag.py
```diff
@@ -1,10 +1,10 @@
 import numpy as np
-
+import json
 from mumaxplus import Ferromagnet, Grid, World, _cpp
 
 
 def demag_field_py(magnet):
-    kernel = _cpp._demag_kernel(magnet._impl)
+    kernel = _cpp._demag_kernel(magnet._impl, 11, -1)
```
lamoreel, line 7: This line (and possible this whole test) can probably be improved since PR #75.

Also: maybe also test a user-set switching radius (small enough so the json files can still be used).

lamoreel (changes requested):

lamoreel (approved):

test/test_demag.py
```diff
@@ -34,12 +36,86 @@ def demag_field_py(magnet):
         (h.shape[3] - mag.shape[3]) :,
     ]
 
-
+@pytest.mark.slow
```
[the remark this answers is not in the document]
ilateur: In the interest of faster tests, I suggest you combine similar magnets and kernels into pytest.fixtures with a class scope. There is no reason to recalculate the same kernel 3 times just to take a different component of it.

```python
@pytest.fixture(scope="class")
def kernel_3d():
    nx, ny, nz = 126, 64, 8
    world = World((1e-9, 1e-9, 1e-9))
    magnet = Ferromagnet(world, Grid((nx, ny, nz)))
    kernel = _cpp._demag_kernel(magnet._impl, 11, 5e-10, -1)
    return kernel
        
class TestDemag:
    def test_Nxx(self, kernel_3d):
        # take slice or whatever
```
They should "only" take 13s combined then, instead of each.

The same can be done for the 100x100 magnets but is less necessary.

ilateur (changes requested):
Sorry this took 2 months...

My biggest concern is how long these calculations tend to take. I already suggested a lot of optimizations, but I fear this won't do that much as there's simply a lot of terms to loop over and sum up on device...

The one test with slightly rectangular cell size aspect ratio and with an accuracy of only 1% is also pretty concerning.

The kernel is also incredibly inaccessible, which sucks to try and change asymptotic parameters in actual simulations (already doable with a recent change on dev but still bad), but this is a refactoring beyond this pull request I (we?) think.

Sorry, something positive: the improved physical accuracy is very nice and the applied math is cool and expandable, very nice in this regard.

mumaxplus/strayfield.py
```diff
@@ -34,3 +34,31 @@ def set_method(self, method):
...
+    def set_epsilon(self, eps):
+        """Set epsilon to calculate the analytical error. The error is defined
+        as epsilon * R³ / V.
+        The default value is 5e-10
+        """
+        self._impl.set_epsilon(eps)
+
+    def set_switch_radius(self, R=-1):
+        """Set the radius from which the asymptotic expantion should be used.
+        Default is -1, then the OOMMF method is used:
```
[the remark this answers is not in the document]
DiegoDeGusem: Is OOMMF error better? Because it is the error method they use

src/physics/demagasymptotic.cu
```diff
@@ -0,0 +1,126 @@
...
+    int N = expansionNxxptr[i];
+    int a = expansionNxxptr[i+1];
+    int b = expansionNxxptr[i+2];
+    int c = expansionNxxptr[i+3];
+    int P = expansionNxxptr[i+4];
+    int d = expansionNxxptr[i+5];
+    int e = expansionNxxptr[i+6];
+    int f = expansionNxxptr[i+7];
+
+    double A = fac(d+2) * fac(e+2) * fac(f+2);
```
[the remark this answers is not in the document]
ilateur: Small enough, but the docstring of set_order should probably warn about a maximum numerical order. Maybe an `assert`

PR #70, opened 2025-04-23
Feature/frozenspins

ilateur:
Add `frozen_spins` parameter like in mumax³. This can be useful for setting custom fixed boundary conditions. It works by setting the calculated torque to (0, 0, 0) if frozenspins is not 0 there.

### DIFF
--- a/mumaxplus/ferromagnet.py
+++ b/mumaxplus/ferromagnet.py
@@ -597,6 +597,15 @@ class Ferromagnet(Magnet):
     def amr_ratio(self, value):
         self.amr_ratio.set(value)
 
+    @property
+    def frozen_spins(self):
+        """Defines spins that should be fixed."""
+        return Parameter(self._impl.frozen_spins)
+    
+    @frozen_spins.setter
+    def frozen_spins(self, value):
+        self.frozen_spins.set(value)
+
     # --- magnetoelasticity ---
 
     @property
--- a/mumaxplus/ferromagnet.py
+++ b/mumaxplus/ferromagnet.py
@@ -599,7 +599,8 @@ class Ferromagnet(Magnet):
 
     @property
     def frozen_spins(self):
-        """Defines spins that should be fixed."""
+        """Defines spins that should be fixed by setting torque to (0, 0, 0)
+        wherever frozen_spins is not 0."""
         return Parameter(self._impl.frozen_spins)
     
     @frozen_spins.setter
--- a/src/physics/torque.cu
+++ b/src/physics/torque.cu
@@ -15,20 +15,22 @@ Field evalTorque(const Ferromagnet* magnet) {
   Field torque = evalLlgTorque(magnet);
   if (!spinTransferTorqueAssuredZero(magnet))
     torque += evalSpinTransferTorque(magnet);
-  freezeSpins(torque, magnet->frozenSpins);
   return torque;
 }
 
 __global__ void k_llgtorque(CuField torque,
                             const CuField mField,
                             const CuField hField,
-                            const CuParameter alpha) {
+                            const CuParameter alpha,
+                            const CuParameter frozenSpins) {
   int idx = blockIdx.x * blockDim.x + threadIdx.x;
 
-  // When outside the geometry, set to zero and return early
-  if (!torque.cellInGeometry(idx)) {
-    if (torque.cellInGrid(idx))
-      torque.setVectorInCell(idx, real3{0, 0, 0});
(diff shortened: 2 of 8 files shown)


mumaxplus/ferromagnet.py
```diff
@@ -597,6 +597,15 @@ def amr_ratio(self):
...
 
+    @property
+    def frozen_spins(self):
+        """Defines spins that should be fixed."""
+        return Parameter(self._impl.frozen_spins)
+    
+    @frozen_spins.setter
+    def frozen_spins(self, value):
+        self.frozen_spins.set(value)
+
```
lamoreel, line 608: I think the naming `frozen_spins` (basically a `bool`) clashes with the rich properties of a `Parameter`. Maybe `scale_torque` is a better name?
Also, this docstring isn't entirely correct. The `Parameter` values scale torque values in the same positions, so it doesn't "define fixed spins".

src/physics/torque.cu
```diff
@@ -51,6 +52,28 @@ Field evalLlgTorque(const Ferromagnet* magnet) {
...
+
+  // When outside the geometry, set to zero and return early
+  if (!torque.cellInGeometry(idx)) {
+    if (torque.cellInGrid(idx))
+      torque.setVectorInCell(idx, real3{0, 0, 0});
+    return;
+  }
+
+  if (frozenSpins.valueAt(idx) != 0)
+    torque.setVectorInCell(idx, real3{0., 0., 0.});
```
lamoreel, line 67: Okay, so whatever you put inside the Python `property`, this only checks whether the value is zero or not. I would add this to the documentation inside `ferromagnet.py`.

test/test_frozenspins.py
```diff
@@ -0,0 +1,51 @@
...
+
+def test_frozenspins_inhomogeneous():
+    """Frozen spins in half the geometry."""
+    world = World((500e-9/128, 125e-9/32, 3e-9/2))
+    magnet = Ferromagnet(world, Grid((128, 32, 2)))
+
+    magnet.msat = 800e3
+    magnet.aex = 13e-12
+    magnet.alpha = 0.02
+    magnet.frozen_spins = 1
```
lamoreel, line 34: This line is already accounted for by your `lambda` function.

test/test_frozenspins.py
```diff
@@ -0,0 +1,51 @@
...
+    # freeze right half
+    center_x = magnet.center[0]
+    magnet.frozen_spins = lambda x, y, z: 1 if x > center_x else 0
+
+    magnet.magnetization = (1, .1, 0)
+
+    # minimize
+    magnet.minimize()
+    tol = 1e-6
+    expectv(magnet.magnetization.average(), (0.9809795618057251, 0.11742201447486877, 0), tol)
```
lamoreel, line 46: Please document where these values come from.

src/physics/torque.cu
```diff
@@ -51,6 +52,28 @@ Field evalLlgTorque(const Ferromagnet* magnet) {
...
+}
+
+void freezeSpins(Field &torque, const Parameter &frozenSpins) {
+  if (!frozenSpins.assuredZero()) {
```
lamoreel, line 71: I would omit this check and do this inside the kernel functions which calculate torques. Since this only looks for zero values, this is basically the same. This way, an extra kernel call (to `k_freezespins`) is removed.

lamoreel (changes requested):

mumaxplus/ferromagnet.py
ilateur: The first implementation was more like a torque scaling, but that seems pretty unphysical. I would keep this boolean-like (properly documented) and let gamma be a user-definable parameter (different issue).

src/physics/torque.cu
ilateur: So pass and use `frozenSpins` in 4 torque functions: `k_llgtorque`, `k_dampingtorque`, `k_ZhangLi` and `k_Slonczewski`?

test/test_frozenspins.py
ilateur: Is a comment such as "make sure behavior does not change as of writing this test" sufficient?

ilateur:
Too many small changes, make sure to squash and merge instead of simply merging.

src/physics/torque.cu
lamoreel: Yes, this check is an early exit.

lamoreel (approved):

PR #73, opened 2025-04-24
Add magnetic force microscopy

DiegoDeGusem:
Now we are capable of creating MFM images for ferromagnets and antiferromagnets! The method is similar to the mumax³ method.
For the MFM kernel we need to calculate d**F**/dz = sum **M** . d²**B**/dz² for every tip position. The sum runs over each cell in the magnet, M is the total magnetization and B is the strayfield from the tip in a cell. The tip has been approximated as 2 monopoles with a charge of +- 1/µ0 and a distance settable by the user via `tipsize` (default 1e-3 m).
In this case the stray field can be calculated as **B** = +-**r** / (µ0 4 pi r³), where **r** is the distance between the pole and the cell.
In order to calculate d²B/dz², there is a `delta` parameter that is used to calculate the strayfield when the tip is at position (x,y,z-`delta`), (x,y,z) and (x,y,z+`delta`).
The height of the tip above the sample can also be set by a user by using the `lift` parameter (default 10e-9).

### DIFF
--- a/mumaxplus/antiferromagnet.py
+++ b/mumaxplus/antiferromagnet.py
@@ -405,3 +405,7 @@ class Antiferromagnet(Magnet):
         enable_elastodynamics, elastic_energy, kinetic_energy
         """
         return ScalarQuantity(_cpp.total_energy(self._impl))
+
+    @property
+    def mfm(self):
+        return FieldQuantity(_cpp.mfm(self._impl))
--- a/mumaxplus/antiferromagnet.py
+++ b/mumaxplus/antiferromagnet.py
@@ -408,4 +408,9 @@ class Antiferromagnet(Magnet):
 
     @property
     def mfm(self):
+        """Magnetic force microscopy.
+        This returns the potential energy (in Joules) of a tip with two monopole
+        charges of +-1/µ0 feels when hovering over the magnet. This can then be
+        used to create an MFM image of the antiferromagnet.
+        """
         return FieldQuantity(_cpp.mfm(self._impl))
--- a/mumaxplus/magnet.py
+++ b/mumaxplus/magnet.py
@@ -554,10 +554,11 @@ class Magnet(ABC):
         return StrayField._from_impl(
                         self._impl.stray_field_from_magnet(source_magnet._impl))
     
-    # --- MFM ---
+    # --- Magnetic force microscopy ---
     @property
     def lift(self):
-        """The hight of the MFM tip above the sample
+        """The height (in meters) of the MFM tip above the xy-surface
+        of the sample. (Default = 10e-9 m)
         
         See Also
         --------
@@ -572,15 +573,16 @@ class Magnet(ABC):
         warn = False
         if self.lift.is_uniform:
             warn = self.lift.uniform_value < 0
-        elif _np.any(self.lift.eval() < 0):
+        elif _np.any(self.lift.eval() < 1e-9):
             warn = True
         
         if warn:
-            warnings.warn("The tip will scratch your sample, lift before it is too late!")
+            warnings.warn("The tip will scratch your sample, increase lift above 1 nm before it is too late!")
         
     @property
     def tipsize(self):
-        """The distance between the two magnetic monopoles in the tip
+        """The distance (in meters) between the two magnetic monopoles in the tip.
+        The tip is perpendicular to the xy-surface of the sample. (default = 1e-3 m)
         
         See Also
         --------
--- a/mumaxplus/magnet.py
+++ b/mumaxplus/magnet.py
@@ -12,6 +12,8 @@ from .scalarquantity import ScalarQuantity
 from .strayfield import StrayField
 from .variable import Variable
 
+import warnings
+
 class Magnet(ABC):
     """A Magnet should never be initialized by the user. It contains no physics.
     Use ``Ferromagnet`` or ``Antiferromagnet`` instead.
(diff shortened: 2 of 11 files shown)


src/physics/mfm.cu
```diff
@@ -0,0 +1,120 @@
...
+
+                            real3 m = magnetization.vectorAt(int3{ix, iy, iz});
+                            real E[3];  // Energy of 3 tip positions
+
+                            // Get 3 different tip heights
+                            for (int i = -1; i <= 1; i++) {
+                                // First pole position and field
+                                real3 R = {x0-x,
+                                           y0-y,
+                                           z0 + z - (lift.valueAt(idx) + i*delta)};
```
lamoreel, line 64: If I understand this nest correctly, `R.z` can be both positive or negative This would imply the tip is sometimes inside the sample. Am I missing something? I admit I'm terribly confused, which is entirely my problem.

src/physics/mfm.cu
```diff
@@ -0,0 +1,120 @@
...
+
+    real dFdz = 0.;
+
+    // Loop over pbc
+    for (int Nx = -pbcRepetitions.x; Nx <= pbcRepetitions.x; Nx++) {
+        real xpbc = Nx * mastergrid.size().x;
+        for (int Ny = -pbcRepetitions.y; Ny <= pbcRepetitions.y; Ny++) {
+            real ypbc = Ny * mastergrid.size().y;
+            for (int Nz = -pbcRepetitions.z; Nz <= pbcRepetitions.z; Nz++) {
+                real zpbc = Nz * mastergrid.size().z;
```
lamoreel, line 46: How physical are PBC in the `z`-direction if there's a tip hanging over the sample? Should this be constrained in some way? I.e. restrict `pbcRepetitions.z` or restrict `lift` if the former is not?

lamoreel, line 46: I think reversing the order of these for-loops might be more efficient.

mumaxplus/magnet.py
```diff
@@ -551,3 +553,41 @@ def stray_field_from_magnet(self, source_magnet: "Magnet"):
         """
         return StrayField._from_impl(
                         self._impl.stray_field_from_magnet(source_magnet._impl))
+    
+    # --- MFM ---
+    @property
+    def lift(self):
+        """The hight of the MFM tip above the sample
```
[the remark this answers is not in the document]
lamoreel: Also, maybe mention that the tip lies orthogonal to the `xy`-plane.

src/physics/mfm.cu
```diff
@@ -0,0 +1,120 @@
...
+        return;
+    
+    // The cell-coordinate of the tip (without lift)
+    const real3 cellsize = kernel.system.cellsize;
+    int3 coo = kernel.system.grid.index2coord(idx);
+    real x0 = coo.x * cellsize.x;
+    real y0 = coo.y * cellsize.y;
+    real z0 = coo.z * cellsize.z;
+
+    real tipCharge = 1/MU0;  // charge at the tip
```
lamoreel, line 28: Initializing `preFactor = 1/(4 * pi * MU0)` avoids doing this many times in the for-loop-nest.

mumaxplus/antiferromagnet.py
```diff
@@ -405,3 +405,7 @@ def total_energy(self):
         enable_elastodynamics, elastic_energy, kinetic_energy
         """
         return ScalarQuantity(_cpp.total_energy(self._impl))
+
+    @property
+    def mfm(self):
```
lamoreel, line 410: I'd add some docstring with unit here (and in `ferromagnet.py`).

lamoreel (changes requested):
Great addition! My biggest comment would be the way we differentiate between FM and AFM (structurally). This seems not that extensible.
Also, I would add this feature to `util/` for several reasons:
- It's literally a utility tool.
- MFM parameters remain uninitialized if not needed (which is a good thing in 99% of all simulations).
- No need to differentiate/duplicate code between the Magnet/FM/AFM/... classes.
This can be an MFM class which takes a `Magnet` in its `__init__` function (accounting for future types of magnets) and has all parameters in the same place. This seems like the most extensible option.

If this would-be class holds the `mfm` `FieldQuantity`, this can still be passed to `solve` to get dynamical MFM footage.

This could be extended in the future to a `postprocessing` (parent) class of some sort where other experimental visualization **tools** can be stored.

src/physics/mfm.cu
```diff
@@ -0,0 +1,120 @@
...
+    Field mfm(magnet->system(), 1, 0.0);
+    const CuField magnetization = fullMagnetizationQuantity(magnet).eval().cu();
+    Grid mastergrid = magnet->world()->mastergrid();
+    int3 pbcRepetitions = magnet->world()->pbcRepetitions();
+    CuParameter lift = magnet->lift.cu();
+    CuParameter tipsize = magnet->tipsize.cu();
+    int ncells = mfm.grid().ncells();
+    cudaLaunch(ncells, k_magneticForceMicroscopy, mfm.cu(), magnetization, mastergrid, pbcRepetitions, lift, tipsize);
+    return mfm;
+}
```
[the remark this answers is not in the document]
DiegoDeGusem: The problem I ran into was that `fullMagnetization` requires a `Ferromagnet` or an `Antiferromagnet`. So it won't compile when giving it a `Magnet`

src/physics/mfm.cu
```diff
@@ -0,0 +1,120 @@
...
+    
+    // The cell-coordinate of the tip (without lift)
+    const real3 cellsize = kernel.system.cellsize;
+    int3 coo = kernel.system.grid.index2coord(idx);
+    real x0 = coo.x * cellsize.x;
+    real y0 = coo.y * cellsize.y;
+    real z0 = coo.z * cellsize.z;
+
+    real tipCharge = 1/MU0;  // charge at the tip
+    real delta = 1e-9;       // tip oscillation, take 2nd derivative over this distance
```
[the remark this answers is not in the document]
DiegoDeGusem: This is the value that was also used in mumax³.

src/physics/mfm.hpp
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+#include "quantityevaluator.hpp"
+
+class Ferromagnet;
```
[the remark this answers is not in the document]
DiegoDeGusem: That's a great question... but yes, it compiled correctly

src/physics/mfm.cu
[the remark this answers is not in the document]
lamoreel: For inspiration, take a look at `evalTotalEnergy` in `energy.cu`. This is a way to overload functions which can take different class instances of sister classes.

src/physics/mfm.cu
DiegoDeGusem: Yes that's right. So, it scans each z-layer and thus once it is done with the top layer it goes "inside" the sample and you can essentially look inside of it. I could make it such that it just doesn't calculate those, but I thought it could be nice/cool to be able to look inside of your sample, but maybe that's strange.

src/physics/mfm.cu
DiegoDeGusem: I don't think PBC in the z-direction are physical in that scenario, but like I commented above, maybe it's fun to look inside a sample

DiegoDeGusem:
Now that I have resolved most of the comments, I will try to make it class in utils :)

ilateur:
This PR has been trumped by PR #77.

Issue #72, opened 2025-04-24
Relax yields NaN when torque is exactly 0


ilateur:
There is no real reason to call relax when there is no torque at all, but still.
It does work when you manually set a relax torque threshold, but not when it changes adaptively.

Example code where it fails:
```python
from mumaxplus import World, Grid, Ferromagnet

length, width, thickness = 500e-9, 125e-9, 3e-9
nx, ny, nz = 128, 32, 1

world = World(cellsize=(length / nx, width / ny, thickness / nz))
magnet = Ferromagnet(world, Grid((nx, ny, nz)))

magnet.msat = 800e3
magnet.aex = 13e-12
magnet.alpha = 0.02
magnet.enable_demag = False
world.bias_magnetic_field = (10e-3, 0, 0)

magnet.magnetization = (1, 0, 0)

magnet.relax()

print(magnet.magnetization.eval())
```

PR #74, opened 2025-04-25
Fix issue #72

lamoreel:
This PR closes issue #72.


ilateur (approved):

ilateur:
Should there be a test which expects the proper error? Similar to `test_set_invalid_method` in `test_timesolver.py`

lamoreel:
Hmm, perhaps, although I think this might be overkill (it's a delicate interplay between magnets in `World` and its associated `TimeSolver`). Not every `std::runtime_error` is caught in a `pytest` test. However, if you want to, a timestep of `zero` could be added to `test_set_invalid_timepoints` in `test_timesolver.py`.

Issue #75, opened 2025-04-28
`StrayField` functionalities are not accessible


lamoreel:
The demag field of a Python `Ferromagnet` is no Python-`StrayField` instance (but a general `FieldQuantity`), so functionalities like `set_method` cannot be used. Internally, the demag field (and stray field) is a C++-`StrayField` instance, but this is not accessible.

PR #76, opened 2025-04-30
Fix issue #75

lamoreel:
Since `StrayField` inherits from `FieldQuantity`, this was an easy fix. The `StrayField` methods of a `Magnet`'s `stray_field` were already accessible.

### DIFF
--- a/mumaxplus/ferromagnet.py
+++ b/mumaxplus/ferromagnet.py
@@ -11,7 +11,6 @@ from .interparameter import InterParameter
 from .parameter import Parameter
 from .poissonsystem import PoissonSystem
 from .scalarquantity import ScalarQuantity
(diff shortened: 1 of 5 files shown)


mumaxplus/ferromagnet.py
ilateur, line 1: What about AFM sublattices? In that case this will crash (`stray_field_from_magnet` won't find anything), but AFM does not have this property now, leaving demag inaccessible.

The previously used `evalDemagField` only worked on ferromagnets, but returned the demag field of the host lattice if asked.

Is the cleanest solution to do the same here (check `is_sublattice`) or to make it a `Magnet` property instead, so it can be asked of AFM, but still crash for sublattices? Or something else?

PR #81, opened 2025-05-13
Add viscous damping to stabilize high frequency oscillations

ilateur:
The phenomenological damping already present tends to dampen rigid body motions, such as translation and rotation, hence it mostly dampens low frequency oscillations. The new viscous damping utilizes a viscosity tensor and strain rate to create a new viscous stress. This tends to dampen high frequency oscillations.

- Add strain rate quantity
- Add viscosity tensor in the form of eta11, eta12 and eta44
- Add viscous stress to the stress tensor
- Rename the old stress tensor to elastic stress
- Rename elastic force to internal body force
  - Elastic and viscous force cannot be (easily) split up because they share the traction-free boundary conditions. The stress is split up though: stress tensor = elastic stress + viscous stress.
- Add formulas to calculate Rayleigh damping coefficients
- Add bulk modulus formula
  - I needed it for a previous version, but not anymore. I figured it could stay, although it has nothing to do with damping anymore.
- Add minuscule test about stress tensor addition
- Add `test_viscousforce_uniform_eta.py`, which exactly mimics `test_elasticforce_uniform_c.py` (same math anyway)
- Add Rayleigh damping to `pure_elastic.py` example

I started working on a Rayleigh damping notebook example, but realized I was just copying information readily available on the internet. The code-part would simply import 2 formulas and apply them to the various eta's. I felt like these few lines were better added to an existing example instead.


mumaxplus/magnet.py
```diff
@@ -403,12 +458,55 @@ def strain_tensor(self):
         rigid_norm_strain, rigid_shear_strain
         """
         return FieldQuantity(_cpp.strain_tensor(self._impl))
+
+    @property
+    def strain_rate(self):
+        """Time derivative of the strain tensor (m/m/s), calculated according to
```
DiegoDeGusem, line 464: m/m/s = 1/s?

src/physics/straintensor.cu
```diff
@@ -114,3 +114,29 @@ Field evalStrainTensor(const Magnet* magnet) {
...
+
+  // The math for strain rate is exactly the same as for strain tensor,
+  // but applied to velocity instead of displacement.
+  cudaLaunch(ncells, k_strainTensor, srField.cu(), v, w, mastergrid);
+
+  return srField;
+}
+
+M_FieldQuantity strainRateQuantity(const Magnet* magnet) {
+  return M_FieldQuantity(magnet, evalStrainRate, 6, "strain_rate", "m/m/s");
```
DiegoDeGusem, line 141: Same comment as before.

test/test_viscousforce_uniform_eta.py
```diff
@@ -0,0 +1,225 @@
...
+        v[v_comp] = A * math.sin(k * relevant_coord)
+        return tuple(v)
+
+    magnet.elastic_velocity = velocity_func
+    
+    force_num = magnet.internal_body_force.eval()
+    force_anal = - k**2 * eta * magnet.elastic_velocity.eval()
+
+    print(force_anal)
+    print(force_num)
```
DiegoDeGusem, line 79: Shouldn't these be removed?

DiegoDeGusem (changes requested):
Nice!
Maybe adding a standard value for the viscous damping would be nice in order to keep the systems from exploding. Or you could say that it's a user error for not using this damping, but that's not very nice.

mumaxplus/magnet.py
```diff
@@ -403,12 +458,55 @@ def strain_tensor(self):
...
+    @property
+    def viscous_stress(self):
+        """Viscous stress tensor (N/m²) due to isotropic viscous damping,
+        calculated according to
+
+        σ = η_b vol(dε/dt) + η_ν dev(dε/dt)
+
+        with vol(dε/dt) and dev(dε/dt) the volumetric and deviatoric parts of
+        the strain rate tensor, and with η_b and η_ν the bulk and shear
+        viscosity respectively.
```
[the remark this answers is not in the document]
ilateur: This is an artifact of the previous implementation and should be changed

ilateur:
Regarding a "default value", this depends on
- cell sizes (set beforehand)
- number of cells (also set beforehand)
- C11, C12 and C44, all set individually, perhaps non-uniform or even time-varying (although that sounds like a bad idea)
- rho, also may be non-uniform (or time-varying)
- some chosen damping ratio, like 1%
- some chosen number of cells in a period, like 4

So in the beginning these will all be 0, then 4 times (rho, C11, C12, C44) when the user sets a parameter, this default value will be recalculated. If the user would like to set viscosity themselves, they should do it after the others, because otherwise it will be overwritten (unless some user-setted check is also implemented).

I would leave it as a formula in the formulary and let the user have common sense. We also don't set a default value for alpha (it's 0) but magnetic simulations can become unstable without it. Nor do we check cell sizes and the user has to manually check the exchange length (also in the formulary) or some other length parameter. I would keep this philosophy.

lamoreel:
> Regarding a "default value", this depends on
> ...

As discussed before, I don't agree with this. Viscosity should be something the user **must be able** to set themselves, but it shouldn't be something the user **has** to set themselves in order to obtain stable simulations. This is not straightforward enough. People who are used to running micromagnetic simulations know that you cannot obtain proper results without a decent `alpha` or `msat`... but viscosity is not trivial.
I suggest calculating _proper default values_ behind the scenes (well documented of course), unless the user sets some themselves. A user-setted check can account for this, or some function which throws a warning (and is only called once). These options should give minimal overhead.

Another option, is excluding the possibility for a zero-viscous damping all together (since this leads to instabilities, this could be okay?) and implementing an `assuredZero` check. Then only calculate default values if the user hasn't set some themselves (which is in 99.999% of cases).

examples/pure_elastic.py
```diff
@@ -58,11 +58,32 @@ def displacement_to_scatter_data(magnet, scale, skip):
...
+magnet.rho = rho
+C11, C44, C12 = 283e9, 58e9, 166e9
+magnet.C11, magnet.C44, magnet.C12 = C11, C44, C12
+
+# --------------------
+# Add Rayleigh damping for stability
+
+# lowest relevant frequency has a wavelength similar to the length of the system
+lowest_freq = np.sqrt(min(C11, C44) / rho) / (2 * max(length, width))
+# highest relevant frequency has a wavelenght of a few cells
```
lamoreel, line 71: wavelength (typo)

mumaxplus/util/formulary.py
```diff
@@ -157,3 +157,96 @@ def magnetic_hardness(K1, msat):
...
+
+
+def Rayleigh_damping_mass_coefficient(frequency_1, damping_ratio_1, frequency_2, damping_ratio_2):
+    """Rayleigh damping mass coefficient calculated by providing damping ratios
+    at specified frequencies.
+
+    4π f₁ f₂ * (ζ₁f₂ - ζ₂f₁) / (f₂² - f₁²)
+    with f₁/f₂ < ζ₂/ζ₁ < f₂/f₁
+
+    Based on https://www.comsol.com/blogs/how-to-model-different-types-of-damping-in-comsol-multiphysics.
```
lamoreel, line 189: The definition of these symbols appear very late in the comsol blog (and only in a UI screenshot). Perhaps these can be explained when defining the different arguments in this docstring?

mumaxplus/util/formulary.py
```diff
@@ -157,3 +157,96 @@ def magnetic_hardness(K1, msat):
...
+    
+    Parameters
+    ----------
+    frequency_1 : float
```
lamoreel, line 193: Please add units of these four parameters.

mumaxplus/util/formulary.py
```diff
@@ -157,3 +157,96 @@ def magnetic_hardness(K1, msat):
...
+    assert damping_ratio_1 >= 0 and damping_ratio_2 >= 0
+    assert frequency_1 >= 0 and frequency_2 >= 0 and frequency_2 > frequency_1
+    assert frequency_1 / frequency_2 < damping_ratio_2 / damping_ratio_1 and \
+           damping_ratio_2 / damping_ratio_1 < frequency_2 / frequency_1
+
+    return 4 * _np.pi * frequency_1 * frequency_2 \
+        * (damping_ratio_1 * frequency_2 - damping_ratio_2 * frequency_1) \
+        / (frequency_2*frequency_2 - frequency_1*frequency_1)
+
+def Rayleigh_damping_stiffness_coefficient(frequency_1, damping_ratio_1, frequency_2, damping_ratio_2):
```
lamoreel, line 218: Same comments about docstring.

mumaxplus/util/formulary.py
```diff
@@ -157,3 +157,96 @@ def magnetic_hardness(K1, msat):
...
+
+    Returns
+    -------
+    float
+        Rayleigh damping mass coefficient (1/s).
+    """
+    assert damping_ratio_1 >= 0 and damping_ratio_2 >= 0
+    assert frequency_1 >= 0 and frequency_2 >= 0 and frequency_2 > frequency_1
+    assert frequency_1 / frequency_2 < damping_ratio_2 / damping_ratio_1 and \
+           damping_ratio_2 / damping_ratio_1 < frequency_2 / frequency_1
```
lamoreel, line 212: If `damping_ratio_1` is set to zero (which is allowed), this will fail, right?

mumaxplus/util/formulary.py
```diff
@@ -157,3 +157,96 @@ def magnetic_hardness(K1, msat):
...
+
+    Returns
+    -------
+    float
+        Rayleigh damping stiffness coefficient (s).
+    """
+    assert damping_ratio_1 >= 0 and damping_ratio_2 >= 0
+    assert frequency_1 >= 0 and frequency_2 >= 0 and frequency_2 > frequency_1
+    assert frequency_1 / frequency_2 < damping_ratio_2 / damping_ratio_1 and \
+           damping_ratio_2 / damping_ratio_1 < frequency_2 / frequency_1
```
lamoreel, line 248: Same concern about `damping_ratio_1 = 0`.

src/physics/straintensor.cu
```diff
@@ -114,3 +114,29 @@ Field evalStrainTensor(const Magnet* magnet) {
 M_FieldQuantity strainTensorQuantity(const Magnet* magnet) {
   return M_FieldQuantity(magnet, evalStrainTensor, 6, "strain_tensor", "");
 }
+
+// --------------------
+// Strain Rate
+
+Field evalStrainRate(const Magnet* magnet) {
+  Field srField(magnet->system(), 6);  // symmetric 3x3 tensor
```
lamoreel, line 122: `srField` is not a transparant name.

src/physics/stresstensor.cu
lamoreel, line 1: I don't like the fact that the structure of this file is entirely different than all other `.cu` files. Maybe reshuffle functions or split into two files.

src/physics/stresstensor.cu
```diff
@@ -52,7 +55,56 @@ Field evalStressTensor(const Magnet* magnet) {
...
+  CuParameter eta44 = magnet->eta44.cu();
+
+  // same mathematics
+  cudaLaunch(ncells, k_elasticStress, stressTensor.cu(), strainRate.cu(), eta11, eta12, eta44);
+  return stressTensor;
+}
+
+M_FieldQuantity viscousStressQuantity(const Magnet* magnet) {
+  return M_FieldQuantity(magnet, evalViscousStress, 6, "viscous_stress", "N/m2");
+}
```
lamoreel, line 95: Does it make more sense to express this unit as `Pa`? :thinking:

src/physics/stresstensor.cu
```diff
@@ -52,7 +55,56 @@ Field evalStressTensor(const Magnet* magnet) {
...
+// --------------------------------------------------
+// Total Stress Tensor
+
+bool stressTensorAssuredZero(const Magnet* magnet) {
+  return elasticityAssuredZero(magnet) && viscousDampingAssuredZero(magnet);
+}
+
+Field evalStressTensor(const Magnet* magnet) {
+  Field stressTensor = evalElasticStress(magnet);  // elastic stress or safely 0
+  if (!viscousDampingAssuredZero(magnet)) stressTensor += evalViscousStress(magnet);
```
lamoreel, line 106: `evalViscousStress` also calls `viscousDampingAssuredZero` which makes this `if` statement unnecessary (but I get the resemblence to `effectivefield.cpp`.

test/test_viscousforce_uniform_eta.py
```diff
@@ -0,0 +1,222 @@
...
+from mumaxplus import Grid, World, Ferromagnet
+
+
+SRTOL = 1e-4
+SRTOL_MIX = 1e-4
+
+cx, cy, cz = 1.5e-9, 2e-9, 2.5e-9
+cellsize = (cx, cy, cz)
+N1, N2 = 128, 256
+P1, P2 = 2, 3  # number of sinus periods
```
lamoreel, line 23: "sine".

test/test_viscousforce_uniform_eta.py
```diff
@@ -0,0 +1,222 @@
...
+    return magnet
+
+
+def set_and_check_sine_force(magnet, d_comp, v_comp, eta):
+    """Sets given component of the velocity of the given magnet to a sine
+    wave, the periodicity of which depends on the direction of the derivative.
+    Then checks if the simulated force corresponds to the expected analytical force.
+    You need to set eta yourself.
+    """
+    magnet.enable_elastodynamics = True  # just in case
```
lamoreel, line 63: No.

test/test_viscousforce_uniform_eta.py
```diff
@@ -0,0 +1,222 @@
...
+    set_and_check_sine_force(magnet, d_comp=2, v_comp=2, eta=eta11)
+
+# ==================================================
+# Tests for double derivate along the different direction
+# eta44 is constant TODO: vary eta44 as well (careful of mixed derivatives!)
+
+def test_dy_dy_vx():
+    magnet = make_long_magnet(d_comp=1)
+    magnet.eta44 = eta44
+    magnet.eta12 = -eta44  # to remove mixed derivate
```
lamoreel, line 107: derivate?

test/test_viscousforce_uniform_eta.py
```diff
@@ -0,0 +1,222 @@
...
+
+def test_dy_dy_vz():
+    magnet = make_long_magnet(d_comp=1)
+    magnet.eta44 = eta44
+    magnet.eta12 = -eta44  # to remove mixed derivate
+    set_and_check_sine_force(magnet, d_comp=1, v_comp=2, eta=eta44)
+
+
+# ==================================================
+# Tests for mixed derivates
```
lamoreel, line 144: derivates?

test/test_viscousforce_uniform_eta.py
```diff
@@ -0,0 +1,222 @@
...
+# eta12 is constant TODO: vary eta12 as well (no eta44 so no double derivative!)
+
+def analytical_mixed_force(k_outer, k_inner, d_comp_outer, d_comp_inner, mgrid):
+    force = np.zeros_like(mgrid)
+    force[d_comp_inner, ...] = A * eta12 * k_outer * k_inner * \
+                                np.cos(k_inner * mgrid[d_comp_inner]) *\
+                                np.cos(k_outer * mgrid[d_comp_outer])
+    return force
+
+def check_mixed_derivative(d_comp_outer, d_comp_inner):
```
lamoreel, line 155: Why here no `set_and_check`?

test/test_viscousforce_uniform_eta.py
```diff
@@ -0,0 +1,222 @@
...
+    """
+    
+    # make world
+    gridsize, pbc_repetitions = [0, 0, 0], [0, 0, 0]
+    gridsize[d_comp_outer], gridsize[d_comp_inner] = N1, N2
+    pbc_repetitions[d_comp_outer], pbc_repetitions[d_comp_inner] = 1, 1
+    world = World(cellsize, mastergrid=Grid(gridsize), pbc_repetitions=pbc_repetitions)
+
+    # make magnet
+    gridsize = [1, 1, 1]  # other index will stay n1
```
lamoreel, line 174: Not clear what this comment means.

lamoreel (changes requested):

mumaxplus/util/formulary.py
ilateur: Good point :thinking:

mumaxplus/util/formulary.py
lamoreel: I feel like either you give a formula and explain it (a bit), or you just reference some source which explains how to calculate the given quantity.
If you would find a better source, then this blog can be replaced. Otherwise, I think this is fine, as long as we reference something (and use correct terminology) since this is only a docstring (or maybe I'm just dense).

src/physics/stresstensor.cu
lamoreel: We adopt our self-made conventional order of
- assuredZero
- kernel functions
- eval functions
- (A)(F)M_FieldQuantities 

and I would prefer to keep this :).

src/physics/stresstensor.cu
ilateur: I personally associate Pa with normal pressure, which is contained in the diagonal of the stress tensor, but Pa for shear stress (force parallel to a surface, instead of perpendicular to it) feels weird to me. But Wikipedia expresses shear stress in Pa too, so...

src/physics/stresstensor.cu
ilateur: I'd rather have a redundant if statement here than a redundant summation with a 0 field :smile: 
The if statement in evalViscousStress could be removed (don't call it if you know it will be zero), but yeah see every magnetic field

test/test_viscousforce_uniform_eta.py
ilateur: It was not a day of spelling well I see

test/test_viscousforce_uniform_eta.py
ilateur: I think it used to mean something sometime, but the code has been changed since the boundary conditions have changed :thinking:

src/physics/stresstensor.cu
ilateur: I think there are multiple .cu files that don't follow the order of the .hpp files :thinking:

mumaxplus/util/formulary.py
ilateur: This is "solved" by asserting <= instead, or by using the much simpler formula, ζ/(πf), if no mass damping is desired

mumaxplus/util/formulary.py
ilateur: Maybe this "source" is a bit clearer?
https://doc.comsol.com/6.3/doc/com.comsol.help.sme/sme_ug_modeling.05.126.html

lamoreel: Also comsol, but more clear.

src/physics/stresstensor.cu
lamoreel: I agree with everything you just said.

ilateur:
I added `stiffness_damping` as an extra Parameter to Magnet. As a default value, this is set to 5e-14 / pi, which corresponds to a damping ratio of 5% at 1 THz or 1% at 0.2 THz. A typical material will have a pressure-wave velocity of ~5500 m/s, so 1 THz corresponds to a wavelength of about 5.5 nm, 0.2 THz to roughly 30 nm. Asuming typical cellsizes of 1 to 5 nm, this is corresponds to a few cells in one wavelength.

The `stiffness_damping` is multiplied by the stiffness tensor components (`C11`, `C12`, `C44`) to replace the effective viscosity tensor inside `viscous_stress` when no viscosity tensor components (`eta11`, `eta12`, `eta44`) have been set. As soon as one is not assuredZero (anywhere, anytime), this `stiffness_damping` is **ignored** and the more general viscosity tensor is used instead. `stiffness_damping` can also be set to 0 if so desired.

mumaxplus/magnet.py
```diff
@@ -311,15 +311,48 @@ def eta(self):
...
+    def stiffness_damping(self, value):
+        self.stiffness_damping.set(value)
+
     @property
     def eta11(self):
         """Viscosity constant eta11 = eta22 = eta33 of the viscosity tensor (Pa s)
         in Voigt notation, which connects strain rate to viscous stress
         σ = η : dε/dt.
+
+        When non-zero somewhere, this parameter overrules the stiffness_damping.
```
lamoreel, line 350: This implies a default value of zero? This is not clear in this docstring. Perhaps say "When set, this parameter..."? (Comment applies to every tensor element.)

examples/pure_elastic.py
```diff
@@ -68,19 +68,17 @@ def displacement_to_scatter_data(magnet, scale, skip):
 
 # lowest relevant frequency has a wavelength similar to the length of the system
```
lamoreel, line 69: I cannot select lines which are not altered (for some reason), but I'd remove the comment saying you add Rayleigh damping **for stability**.
Also, should you add Rayleigh damping in such a basic example? This is accounted for automatically now, so may be overkill to show the user everything at once.

mumaxplus/util/formulary.py
```diff
@@ -179,74 +179,83 @@ def bulk_modulus(C11, C44):
...
+    See Also
+    --------
+    Rayleigh_damping_stiffness_coefficient
     """
     assert damping_ratio_1 >= 0 and damping_ratio_2 >= 0
     assert frequency_1 >= 0 and frequency_2 >= 0 and frequency_2 > frequency_1
-    assert frequency_1 / frequency_2 < damping_ratio_2 / damping_ratio_1 and \
-           damping_ratio_2 / damping_ratio_1 < frequency_2 / frequency_1
+    assert frequency_1 / frequency_2 <= damping_ratio_2 / damping_ratio_1 and \
+           damping_ratio_2 / damping_ratio_1 <= frequency_2 / frequency_1
```
lamoreel, line 218: `damping_ratio_1 = 0` still causes trouble, right?

mumaxplus/util/formulary.py
```diff
@@ -179,74 +179,83 @@ def bulk_modulus(C11, C44):
...
-    assert damping_ratio_1 >= 0 and damping_ratio_2 >= 0
-    assert frequency_1 >= 0 and frequency_2 >= 0 and frequency_2 > frequency_1
-    assert frequency_1 / frequency_2 < damping_ratio_2 / damping_ratio_1 and \
-           damping_ratio_2 / damping_ratio_1 < frequency_2 / frequency_1
+    assert damping_ratio >= 0 and frequency >= 0
 
-    return 1 / _np.pi \
-        * (damping_ratio_2 * frequency_2 - damping_ratio_1 * frequency_1) \
-        / (frequency_2*frequency_2 - frequency_1*frequency_1)
+    return damping_ratio / (_np.pi * frequency)
```
lamoreel, line 261: Is `frequency = 0` allowed?

src/physics/magnet.hpp
```diff
@@ -82,6 +82,7 @@ class Magnet {
   Parameter C44;  // C44 = C55 = C66
 
   Parameter eta;  // Phenomenological elastic damping constant
+  Parameter stiffnessDamping;  // eta_ij = coeff * Cij
```
lamoreel, line 85: Comment doesn't make it clear that this `Parameter` is `coeff`.

src/physics/stresstensor.cu
```diff
@@ -67,11 +67,35 @@ M_FieldQuantity elasticStressQuantity(const Magnet* magnet) {
...
+                                         const CuParameter a,
+                                         const CuField x) {
+  int idx = blockIdx.x * blockDim.x + threadIdx.x;
+  if (!y.cellInGeometry(idx))
+    return;
+  for (int c = 0; c < y.ncomp; c++) {
+    y.setValueInCell(idx, c, a.valueAt(idx) * x.valueAt(idx, c));
+  }
+}
+
```
lamoreel, line 98: I'm getting it out there once again, but I don't think that files inside `physics/` are the best place for such functions.

src/physics/stresstensor.cu
```diff
@@ -81,12 +105,30 @@ Field evalViscousStress(const Magnet* magnet) {
...
+  // use viscosity tensor if set
+  if (!viscousDampingTensorAssuredZero(magnet)) {
+    CuParameter eta11 = magnet->eta11.cu();
+    CuParameter eta12 = magnet->eta12.cu();
+    CuParameter eta44 = magnet->eta44.cu();
+
+    // same mathematics
+    cudaLaunch(ncells, k_elasticStress, stressTensor.cu(), strainRate.cu(), eta11, eta12, eta44);
+
+  } else {  // or use stiffness tensor multiplied by rayleigh damping stiffness coefficient
```
lamoreel, line 118: I'd switch the `if` and `else` scopes (and condition, of course) :).

src/physics/stresstensor.cu
lamoreel: `multiply` in `fieldops.cu` should do this trick, no?

src/physics/stresstensor.cu
ilateur: so like this?
```cpp
if viscousDampingTensorAssuredZero(magnet) {
  // rayleigh damping
} else {
  // tensor damping
}
```
I did it the other way because in my mind "tensor takes priority, so it is calculated first if possible"

src/physics/stresstensor.cu
ilateur: Should I generalize this to fieldOps? That felt a bit overkill, but I can

mumaxplus/util/formulary.py
ilateur: No. Viscous damping has 0 effect at 0 Hz, then increases linearly with frequency. Maybe I should add this to the docstring and assert it.

mumaxplus/util/formulary.py
ilateur: ah... right :thinking: but not anymore when frequency_1 = 0 :grin: then it is simpler to use only `Rayleigh_damping_stiffness_coefficient`.

Explicitly setting 0 damping at non-zero frequency is not possible with A/x + Bx (A,B > 0)

examples/pure_elastic.py
ilateur: Fair point, but I wanted to show it *somewhere*... I thought about commenting something like "viscous damping is enabled by default, but we set it explicitly here" but then forgot

src/physics/stresstensor.cu
lamoreel: Why don't you just get the `Field` values of the `Parameter` and use `multiply` directly? Or maybe this copy is too expensive.

ilateur: Ah yes I could call `Parameter.eval()`, then use that `Field` in the multiplication, but that uses more kernel calls. What's the cleanest option in your opinion?

lamoreel: I think it's more clean to use the already implemented `fieldops`, but that's less efficient...

mumaxplus/magnet.py
ilateur: I wanted to avoid the case where the user sets it to 0, which does nothing, but might give the impression of removing damping.

mumaxplus/magnet.py
```diff
@@ -311,15 +311,48 @@ def eta(self):
...
+        This corresponds to a damping ratio of 5% at a frequency of 1 THz.
+        This is meant to stabilize the high frequency modes, with a typical
+        wavelength of only a few cellsizes in length.
+
+        The viscosity tensor η is assumed to be proportional to the stiffness
+        tensor C, with this coefficient β as the proportionality constant.
+
+        η = β * C
+
+        This parameter is completely IGNORED when any component of the viscosity
```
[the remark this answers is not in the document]
ilateur: That feels a bit extreme

DiegoDeGusem (approved):

PR #86, opened 2025-06-19
Very fast demag

DiegoDeGusem:
This version of the asymptotic demag has all the formulas written out. It is 100 times faster (like literally). However, the build takes a bit longer.


test/test_demag.py
```diff
@@ -34,12 +42,101 @@ def demag_field_py(magnet):
         (h.shape[3] - mag.shape[3]) :,
     ]
 
-
 class TestDemag:
+    def setup_class(self):
```
ilateur, line 46: Because of the speed, these things can be moved back to their relevant tests. But keeping it here is faster and doesn't hurt I suppose (unless someone wants to test only Nxx_3D a lot of times...)

test/test_demag.py
```diff
@@ -34,12 +42,101 @@ def demag_field_py(magnet):
...
+            with the BigFloat package with an accuracy of 1024 bits
+            and the analytical method."""
+        
+        nx, ny, nz = 126, 64, 8
+        world = World((1e-9, 1e-9, 1e-9))
+        magnet = Ferromagnet(world, Grid((nx, ny, nz)))
+        mumaxplus_result = _cpp._demag_kernel(magnet._impl, 11, 5e-10, 5e-9)[0,nz:,ny:, nx:] # Nxx component
+
+        # avoid fake errors when both values are super small
+        mask = ~((np.abs(self.exact_Nxx) < 5e-15) & (np.abs(mumaxplus_result) < 5e-15))
```
ilateur, line 93: Only done in this test?

test/test_demag.py
```diff
@@ -34,12 +42,101 @@ def demag_field_py(magnet):
...
+
+    def test_Nxx_aspect(self):
+        """ Compare the demagkernel with high accurate json files. These were made
+            with the BigFloat package with an accuracy of 1024 bits
+            and the analytical method."""
+        
+        mumaxplus_result = self.kernel_aspect[0,:,ny_aspect:, nx_aspect:] # Nxx component
+
+        err = np.max(relative_error(mumaxplus_result, self.exact_aspect_Nxx))
+        assert err < 1e-2
```
ilateur, line 132: still haven't figured out the reason for this? I find it slightly worrying, but maybe it's "fake errors"?

test/exact_Nxx_3D.json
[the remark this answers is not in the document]
DiegoDeGusem: Yes, I can't compress them significantly. The only way to compress them is by removing spaces and enters, but that does not change the size of the files

test/test_demag.py
DiegoDeGusem: Yes, the normal Nxx was the only test where we divided by things like 1e-305

PR #88, opened 2025-07-07
Energy calculation (bug fix and improvement)

lamoreel:
This PR

- Fixes the energy calculation for non-trivial geometries.
- Renders the energy calculation from an energy density less bug-prone.
- Let `System` calculate `cellsInGeo` only once at initialization.


DiegoDeGusem (approved):
This all seems correct to me

ilateur (approved):

PR #97, opened 2025-09-18
Fix #93: Compiling on Windows

### LINKED ISSUE #93: Compiling on Windows (Successful! -ish)
Hi,

I'm happy to say I was able to get mumax+ to compile on Windows with some tweaking, following the steps given on the github main pages. There are ~4-5 tweaks. I am still running into one final issue/kludge.

Currently, I can get the cmd.exe executable to run without issue, which has the Standardproblem4 and Spinwave Dispersion example problems. It seems to give full results in the csv output file. For the .pyd file, I am able to complete the compilation process with no errors with fixes 1-4 listed below. However, whenever I try to actually import it (say, running the anisotropy.py example), I get a DLL error:

Image

It seems to be related to the CUDA DLLs. If I copy curand64_10.dll and cufft64_11.dll directly into the folder with the .pyd file, it works (I get the output image from say, anisotropy.py). I'm not sure why I need to do this, because they are definitely in my PATH (I also have a copy of them in C:/cuda, due to mumax3):

Image

This was difficult to troubleshoot since there was no error message, but the only other DLLs (that didn't involve things like the kernel, which i did not want to touch) were vcruntime140_1.dll and msvcp140.dll. Copying those alone did not affect the error.

--
The changes I've had to make to the build process so far (these are in order that I got the error messages while compiling):

1. Initial error in mumaxplus\src\core\inter_parameter.cu : Complains about std:string not being imported. Fixed by adding #include <string> into inter_parameter.cu

The same error/fix for mumaxplus\src\core\system.cpp (although this error message will not show until some of the other fixes are made first)

2.  In mumaxplus\src\core\fieldops.cu, it wanted the value for pi to be explicitly declared as a device variable. Insead of `const float pi = 3.1415926535897931f;`  , I changed it to be: `__device__ const float pi = 3.1415926535897931f;`

3. In mumaxplus\src\bindings\wrap_field.cpp and mumaxplus\src\bindings\wrap_voronoi.cpp, the ssize_t type gives an issue on Windows. This is native to Linux/Posix. I fixed this by adding to the top of the file

```
#ifdef _MSC_VER  
  #include <BaseTsd.h>
  typedef SSIZE_T ssize_t;
#endif
```

This is probably not a very good fix, but sufficient just to get things to compile. If any of these changes get merged, there is probably a more elegant way to do this.

4. During the compilation process (if using the conda method), it gave some linker warnings about using /MD vs /MT. This seems to be mostly tied to the cmd  executable (if using the setup.py method, it does not seem to build cmd). 

Image

I could not find a way to fix this warning by editing the root Cmake.txt . I had to edit the Cmake in mumaxplus\src\cmd with:

target_compile_options(cmd PRIVATE /MD)
target_link_options(cmd PRIVATE /NODEFAULTLIB:LIBCMT)

I also had to run Cmake manually, and force a release instead of dev build (using: 
```
mkdir build
cd build
cmake .. -DCMAKE_BUILD_TYPE=Release
cmake --build . --config Release
cmake --install .
```
It's possible this can be avoided using something like /MDd (the development equivalent of /MD) instead of /MD.

5. As mentioned above, copy the CUDA DLLs into the folder with the .pyd (in my case, it seemed to put it in GitHub\mumaxplus by default)

-- 

Benchmark results for my 1070 ti:
   ncells   walltime   throughput
   4194304   20.78604    2.018E+07

For mumax3's bench, at 2048x2048=4194304 cells: 4.194304e+06 1.6878067856423846e+08 1.694520647736488e-14

[benchmark.txt](https://github.com/user-attachments/files/21829425/benchmark.txt)
[benchmarkmumaxplus.txt](https://github.com/user-attachments/files/21829426/benchmarkmumaxplus.txt)

Best regards,
Josh Lauzier

JonathanMaes:
Thanks to the insight of @jplauzie in issue #93 (most importantly, the mysterious missing DLLs turned out to be `cufft64_*.dll` and `curand64_*.dll`), compilation on Windows should now be possible without users having to edit the source code before compiling.

More specifically, this PR implements points 1, 2, 3 and 5 of issue #93. Suggestion 4 (about linker warnings) was skipped for now, since the warning is not a fatal error, and fixing them breaks the compilation process for me.

I also updated the relevant `README.md` section concerning compilation on Windows. It seems the commands to be used are now mostly identical for both Linux and Windows, so a simplification of `README.md` remains possible.

$$\rightarrow$$ Could the reviewers check whether this branch still successfully compiles on Linux systems?

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ All demonstrations in the paper were simulated using version [v1.1.0](https://gi
 You should install these yourself
 * CUDA Toolkit 10.0 or later
 * A C++ compiler which supports C++17, such as GCC
-* On Windows (good luck): MSVC 2019
+* On Windows: MSVC 2019 or 2022
 
 These will be installed automatically within the conda environment
 * cmake 4.0.0
@@ -70,29 +70,21 @@ add_definitions(-DFP_PRECISION=DOUBLE) # FP_PRECISION should be SINGLE or DOUBLE
 
 ### Windows
 
-**These instructions are old and worked at some point (2021), but not today. If you are brave enough to try Windows and you manage to get it working, please let us know!**
-
-1. Install Visual Studio 2019 and the desktop development with C++ workload
-2. Install CUDA Toolkit 10.x
-3. Install cmake
-4. Download the pybind11 submodule with git
-```
+1. Install Microsoft Visual Studio with the "Desktop development with C++" workload
+2. Install the [CUDA Toolkit](https://developer.nvidia.com/cuda-toolkit-archive) *(version 10.0 or later)*
+3. Download the pybind11 submodule with git:
+```powershell
 git submodule init
 git submodule update
 ```
-5. Install Python packages using conda
-```
+4. Install Python packages using conda:
+```powershell
 conda env create -f environment.yml
 ```
-6. Build `mumaxplus` using `setuptools`
-```
-activate mumaxplus
-python setup.py develop
-```
-or `conda`
-```
+5. Build `mumaxplus` using `pip` and `setuptools`:
+```powershell
 conda activate mumaxplus
-conda develop -b .
+pip install -v .
 ```
 
 ## Documentation
--- a/src/bindings/wrap_field.cpp
+++ b/src/bindings/wrap_field.cpp
@@ -7,6 +7,11 @@
 #include "field.hpp"
 #include "wrappers.hpp"
 
+#ifdef _MSC_VER  
+  #include <BaseTsd.h>
+  typedef SSIZE_T ssize_t;
+#endif
+
 void wrap_field(py::module& m) {
   py::class_<Field>(m, "Field")
       .def_property_readonly("grid", &Field::grid)
--- a/src/bindings/wrap_voronoi.cpp
+++ b/src/bindings/wrap_voronoi.cpp
@@ -7,6 +7,10 @@
 #include "voronoi.hpp"
 #include "wrappers.hpp"
 
+#ifdef _MSC_VER  
+  #include <BaseTsd.h>
+  typedef SSIZE_T ssize_t;
+#endif
 
 void wrap_voronoi(py::module& m) {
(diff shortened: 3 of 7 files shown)


lamoreel (changes requested):
Great to finally see this happening..!

README.md
```diff
@@ -1,24 +1,81 @@
...
+## Installation
+
+### Dependencies
+
+You should install the following tools yourself. Click the arrows for more details.
+
+<details><summary>CUDA Toolkit <i>(version 10.0 or later)</i></summary>
+
+* **Windows**: Download an installer from [the CUDA website](https://developer.nvidia.com/cuda-toolkit-archive).
+* **Linux**: Use `sudo apt-get install nvidia-cuda-toolkit`, or [download an installer](https://developer.nvidia.com/cuda-toolkit-archive).
```
ilateur, line 27: *Technically* this depends on your installed package manager (here `apt`), which mostly depends on your Linux distribution. But people who installed another probably already know this...

README.md
```diff
@@ -27,95 +84,56 @@ These will be installed automatically within the conda environment
...
+> [!TIP]
+> The source code can also be compiled with double precision, by changing `FP_PRECISION` in `CMakeLists.txt` from `SINGLE` to `DOUBLE` before rebuilding.
+>
+> ```cmake
+> add_definitions(-DFP_PRECISION=DOUBLE) # FP_PRECISION > should be SINGLE or DOUBLE
+> ```
 
 ## Documentation
 
 Documentation for mumax⁺ can be found at http://mumax.github.io/plus.
```
ilateur, line 114: I wonder if the documentation should (also?) be mentioned before Installation?

DiegoDeGusem (approved):

README.md
lamoreel: Yes, I always have the feeling that this section comes too late, but I also get why it's there...

lamoreel (approved):

lamoreel:
This should be tested on another Windows system before merging.

PR #119, opened 2026-01-08
Promote `GAMMALL` to `Parameter` class

lamoreel:
This PR promotes the gyromagnetic ratio (previously known as `GAMMALL`) to the `Parameter` class. This is now a settable property of `Ferromagnet`, allowing for different values in different sublattices (as can be the case in ferrimagnets) along with all the other `Parameter` functionalities.


test/test_gamma.py
```diff
@@ -0,0 +1,71 @@
...
+        magnet.aex = 13e-12
+    
+    magnet2.magnetization = magnet1.magnetization()
+    magnet2.gamma = magnet1.gamma() * 2
+
+    t1 = magnet1.torque()
+    t2 = magnet2.torque()
+    
+    err = max_relative_error(t2 / 2, t1)
+    assert err < 5e-4
```
ilateur: would `max_semi_relative_error` be better? :thinking: same with other tests

ilateur (commented):
I don't really like that almost 20 of the changed files are to remove a variable and replace it with a copy-pasted number, but otherwise the actual implementation looks fine :)

JonathanMaes (approved):

DiegoDeGusem (approved):

ilateur (approved):

Issue #117, opened 2026-01-07
It is not possible to set different magnetization for different regions.


nikprodev:
Class "Variable" does not have a method "set_in_region" like class "Parameter".

lamoreel:
This feature is now added in PR #121 and PR #122 . This will be included in the next version release of mumax+. In the meantime, you can compile the `dev` branch.

PR #121, opened 2026-01-14
Set `Variable` value per region

lamoreel:
This PR allows for region-wise setting of `Variable` values (as was already possible for `(Vector)Parameter`). This partially answers the request in issue #117 .
**Note:** the possibility to use a function or array to set a non-uniform and/or dynamic value per region is **not included here**. This is a more general `Field`-related issue and will be taken care of later in a separate PR.

### DIFF
--- a/mumaxplus/variable.py
+++ b/mumaxplus/variable.py
@@ -27,6 +27,15 @@ class Variable(FieldQuantity):
         else:
             self._impl.set(value)
 
+    def set_in_region(self, region_idx, value):
+        """
+        Set a uniform, static value in a specified region.
+        """
+        assert (isinstance(value, (float, int)) or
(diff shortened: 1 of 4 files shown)


JonathanMaes (approved):

DiegoDeGusem (approved):

ilateur (approved):
I'll suppose that this works, as I can't test it anymore. Should there be a (dead-simple) unit test about this functionality, a simple set and check?

lamoreel:
> I'll suppose that this works, as I can't test it anymore. Should there be a (dead-simple) unit test about this functionality, a simple set and check?

There's another PR in the making where you can set `Parameter` and `Variable` values using a function. I will include this there.

PR #122, opened 2026-01-19
Feature: Use a function to set `Parameter` and `Variable` values in a single region

lamoreel:
This PR answers to the request in issue #117 (complementary to PR #121), allowing for the use of functions to set static values in `Parameter` and `Variable` regions.

### DIFF
--- a/mumaxplus/variable.py
+++ b/mumaxplus/variable.py
@@ -21,6 +21,10 @@ class Variable(FieldQuantity):
             Or the new value can be set cell by cell with an ndarray with the same
             shape as this variable, or with a function which returns the cell value
             as a function of the position.
+
+        See Also
+        --------
+        :func:`set_in_region`
         """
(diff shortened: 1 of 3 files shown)


mumaxplus/variable.py
ilateur, line 1: It's a bit sad to copy the code for Parameter and Variable, but I suppose it does not really fit with the parent class FieldQuantity?

test/test_set_in_region.py
ilateur, line 1: clean

ilateur (approved):

mumaxplus/variable.py
lamoreel: Not really, since `FieldQuantity` is the abstract base class (and doesn't even have a setter).

DiegoDeGusem (approved):

JonathanMaes (approved):

PR #123, opened 2026-01-19
Move choice of floating-point precision to the Python layer

JonathanMaes:
This PR simplifies the process of switching between single and double floating-point precision.
Previously, the precision had to be chosen before compilation, by editing `CMakeLists.txt`.
Now, mumax+ compiles two binaries: one for single and one for double precision (unless `MUMAXPLUS_FP_PRECISION` is set, see below). The appropriate binary is then loaded during the `import mumaxplus` statement.

Two switches control the floating-point precision, as listed below in descending order of priority. Both can take the values `SINGLE`/`1`/`32` for single precision and `DOUBLE`/`2`/`64` for double precision.
- A command-line flag `--mumaxplus-fp-precision` passed to the `python` command.
  - Example: `python some_mumax_script.py --mumaxplus-fp-precision DOUBLE`.
- The environment variable `MUMAXPLUS_FP_PRECISION`.
  - Setting this will also affect the compilation: mumax+ will then only be compiled for one kind of floating point precision. This can be useful to reduce compile time during development, but passing the opposite precision with `--mumaxplus-fp-precision` will then result in a `ModuleNotFoundError`.

The used precision can then be accessed as `mumaxplus.FP_PRECISION`.

> [!NOTE]
> Since the choice of precision must be made before the `import mumaxplus` statement, it is not possible to use single precision in one part of a script and double precision in another: these must be run in separate processes. Reloading mumax+ with a different floating point precision during runtime may lead to a vast wealth of errors.

The unit test is based on `examples/errorscaling.py`, since the error scaling is highly dependent on the floating point precision.


docs/install.rst
ilateur, line 1: Are these two paragraphs enough clarification, or should the documentation have its own page about setting precision? This (small) page could include for example the choice of single or double precision, the values 2 64 DOUBLE, before or after compilation, flag or environment variable, error scaling example plots, potential errors, etc.
(This pull request description is more clear/complete  than the installation instructions methinks)

mumaxplus/__init__.py
```diff
@@ -20,6 +49,8 @@
 from .world import World
 from . import util
 
+FP_PRECISION = {1: "SINGLE", 2: "DOUBLE"}.get(_cpp.FP_PRECISION, "UNKNOWN") # Use _cpp value, as that is certainly the correct one
```
ilateur, line 61: but you just loaded `_cpp` according to the previous `FP_PRECISION`?

mumaxplus/__init__.py
```diff
@@ -39,4 +70,5 @@
     "World",
     "PoissonSystem",
     "util",
+    "FP_PRECISION"
```
ilateur, line 82: does this mean you can `import` the value `FP_PRECISION` inside scripts?

test/test_fp_precision.py
```diff
@@ -0,0 +1,95 @@
...
+    error = np.linalg.norm(exact - output)
+
+    return error
+
+
+def run_with_precision(precision):
+    FP_PRECISION_CPP = {"SINGLE": 1,
+                        "DOUBLE": 2}.get(precision)
+    FP_PRECISION = {"SINGLE": '"SINGLE"',
+                    "DOUBLE": '"DOUBLE"'}.get(precision)
```
ilateur: this seems silly

test/test_fp_precision.py
ilateur, line 1: It is cool that you managed to run both versions within one file :+1:

ilateur (commented):
Someone (or some-multiple) should test this on linux

mumaxplus/__init__.py
JonathanMaes: Yes, this just ensures `mumaxplus.FP_PRECISION` is either `"SINGLE"` or `"DOUBLE"`, not `1`, `2`, `32` or `64`. What better way to do so than to use the precision of the loaded `_cpp` module to be absolutely sure `FP_PRECISION` represents the used precision?

mumaxplus/__init__.py
JonathanMaes: Indeed. I will add this to the description.

test/test_fp_precision.py
JonathanMaes: that's because it is

docs/install.rst
JonathanMaes: This could be a good idea, if the others think so as well.

DiegoDeGusem (commented):
On my linux device the code compiles and the test succeeds :tada:
I also agree with @ilateur's comment about a larger description.

JonathanMaes:
A webpage was added in the Tutorials section, explaining how to choose a precision, what errors can occur, and showcasing the effect on the error floor.

README.md
lamoreel, line 1: This section is now no longer part of "Building mumax+", but is more related to executing python scripts. Perhaps this can be moved to its own subsection under the same header (after "Troubleshooting" and before "Documentation")?

lamoreel: Also, this section could benefit from a hyperlink to the new `precision.ipynb` tutorial.

examples/precision.ipynb
```diff
@@ -0,0 +1,395 @@
...
+    "  os.environ[\"MUMAXPLUS_FP_PRECISION\"] = \"DOUBLE\"\n",
+    "  ```\n",
+    "  **before** the `import mumaxplus` statement.\n",
+    "\n",
+    "The precision ultimately used by mumax⁺ can be accessed as `mumaxplus.FP_PRECISION`, which will be either `\"SINGLE\"` or `\"DOUBLE\"`.\n",
+    "\n",
+    "### Compilation\n",
+    "\n",
+    "Compiling mumax⁺ normally results in two binaries: one for single and one for double precision.\n",
+    "However, if the environment variable `MUMAXPLUS_FP_PRECISION` is set, only one binary will be produced, corresponding to desired precision.\n",
```
lamoreel: Missing "the" before "desired".

examples/precision.ipynb
```diff
@@ -0,0 +1,395 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "cc3c1a35",
+   "metadata": {},
+   "source": [
+    "## Example: error floor\n",
+    "\n",
+    "The floating-point precision used has a significant impact on the error floor.\n",
+    "\n",
+    "Below, we compare the numerical and analytical result of 10 precessions of one spin in an external magnetic field of 0.1 T, with damping. Depending on the time step, this results in a different error. The minimum attainable error is dependent on the floating-point precision used.\n",
```
lamoreel: Maybe reference the already existing `errorscaling.py`.

examples/precision.ipynb
```diff
@@ -0,0 +1,395 @@
...
+     },
+     "execution_count": 3,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "# Restart the kernel so we can load mumax+ again with a different precision.\n",
+    "# This likely gives a popup saying that the kernel was stopped or will restart.\n",
+    "get_ipython().kernel.do_shutdown(restart=True)"
```
lamoreel, line 195: This doesn't work for me 😕 . (The kernel dies, but refuses to restart when trying to execute the next cell).

```python
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
Cell In[1], [line 8]
        5 import numpy as np
        6 from math import acos, atan, pi, exp, tan, sin, cos, sqrt
  ----> 8 from mumaxplus import *
        9 from mumaxplus.util import *
       12 def magnetic_moment_precession(time, initial_magnetization, hfield_z, damping):

ModuleNotFoundError: No module named 'mumaxplus'
```

mumaxplus/__init__.py
```diff
@@ -1,7 +1,36 @@
...
+# Was a command line argument passed?
+parser = argparse.ArgumentParser()
+parser.add_argument('--mumaxplus-fp-precision', dest='fp_precision', type=str, nargs='?', default=None,
+                    help='Let mumax+ use single (FP_PRECISION = SINGLE, 1 or 32) or double (DOUBLE, 2 or 64) precision. This argument takes precedence over the environment variable MUMAXPLUS_FP_PRECISION.')
+args, unknown = parser.parse_known_args()
+FP_PRECISION = args.fp_precision # Can be None
+
+# If not, was an environment variable set?
+if not FP_PRECISION:
+    FP_PRECISION = os.environ.get("MUMAXPLUS_FP_PRECISION")
```
lamoreel: Could this benefit from an extra check which compares the values set by both the flag and the env var? This could throw a warning if both are set, but differ.

lamoreel (changes requested):
I was able to successfully compile with a single (no pun intended) and with both precisions on a Linux machine 🌈 .
I was also able to choose the precision to be used by both the flag and the env var, hence my pending approval.

Most of my comments are about documentation

README.md
JonathanMaes: Indeed it seems I forgot to update the README.

examples/precision.ipynb
JonathanMaes: Looks like the kernel restarts, but not in the right conda env? The imports before `import mumaxplus` work just fine.

docs/images/fp_precision_single.png
ilateur, line 1: a bit sad that the lines are above the linear scaling regime, because the fit domain is too large... but also eh

test/test_fp_precision.py
ilateur, line 1: cool!

ilateur (approved):

docs/images/fp_precision_single.png
DiegoDeGusem, line 1: I would maybe remove the fits from these plots (and code) as they don't really matter here and they are next to the data points?

DiegoDeGusem (changes requested):

docs/images/fp_precision_single.png
JonathanMaes: Then it looks kinda empty. I moved them to look more acceptable.

JonathanMaes: I moved them to look more acceptable.

ilateur (approved):

DiegoDeGusem (approved):

lamoreel (approved):
The tutorial hyperlink in the README leads to a 404 error, but I'm guessing the website just needs rebuilding after the merge?

JonathanMaes:
Indeed, after rebuilding the website this link should begin to exist.

PR #133, opened 2026-03-24
Add Line and Line2D shapes

ilateur:
This mirrors the recent [pull request of mumax3](https://github.com/mumax/3/pull/383), except `diam=0` is not allowed, as `shape.py` has no concept of grids and cellsizes (and it is cleaner that way anyway).

image


JonathanMaes (approved):
Even with some optimizations that I forgot to put into the mumax3 version :O

This must be that infamous "plus" mindset.

PR #134, opened 2026-03-26
Add option to avoid storing timesolver.solve data in RAM

lamoreel:
This is a follow-up pull request after PR #55.

### DIFF
--- a/mumaxplus/timesolver.py
+++ b/mumaxplus/timesolver.py
@@ -20,7 +20,8 @@ class TimeSolverOutput:
             Specify wether to store the data in an output dictionary.
         """
         self._quantities = quantity_dict
-        self._data = {"time": []}
+        self._values = {}  # dictionary to keep calculated quantities of one step
+        self._data = {"time": []}  # dict to keep all data
         for key in self._quantities.keys():
             self._data[key] = []
 
@@ -42,19 +43,19 @@ class TimeSolverOutput:
         """Compute all the specified quantities for the current state."""
 
         # Compute values
-        values = {"time": time}
+        self._values["time"] = time
         for key, func in self._quantities.items():
-            values[key] = func()
+            self._values[key] = func()
 
         # Store in memory
         if self._store_as_dict:
-            for key, value in values.items():
+            for key, value in self._values.items():
                 self._data[key].append(value)
 
         # Write to file
         if self._file_name is not None:
             with open(self._file_name, 'a') as file:
-                print(*[values[key] for key in self._keys], sep="\t", file=file)
+                print(*[self._values[key] for key in self._keys], sep="\t", file=file)
 
     def __getitem__(self, key):
         """Return the computed values of a quantity."""
--- a/mumaxplus/timesolver.py
+++ b/mumaxplus/timesolver.py
@@ -2,10 +2,11 @@
 
 from typing import Callable
 import os as _os
+import warnings
 from tqdm import tqdm as _tqdm
 
 class TimeSolverOutput:
-    def __init__(self, quantity_dict, file_name=None):
+    def __init__(self, quantity_dict, file_name=None, store_data=True):
         """Collect values of a list of quantities on specified timepoints.
 
         Parameters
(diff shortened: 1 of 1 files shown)


mumaxplus/timesolver.py
```diff
@@ -126,20 +142,23 @@ def solve(self, timepoints, quantity_dict, file_name=None, tqdm=False) -> TimeSo
         file_name : str, optional
             Optional name of an output file, in which the data is also written
             as tab-separated values during the simulation.
+        store_data : bool (default=True)
```
ilateur: I find this name slightly confusing, as writing to a file is also a way of storing data. Perhaps it can be renamed to `store_as_dict` or `keep_in_memory` or something?

mumaxplus/timesolver.py
```diff
@@ -126,20 +142,23 @@ def solve(self, timepoints, quantity_dict, file_name=None, tqdm=False) -> TimeSo
...
+            Specify wether to store the calculated quantities in `quantity_dict`.
```
ilateur: *whether

ilateur: Maybe I am reading this wrong, but this sounds like the data is stored inside of `quantity_dict`, but that only keeps track of quantities which need to be evaluated. The data is stored in `output` (a `TimeSolverOutput` instance).
Maybe this?
> Specify whether to store the evaluated quantities in a dictionary in memory.

mumaxplus/timesolver.py
```diff
@@ -126,20 +142,23 @@ def solve(self, timepoints, quantity_dict, file_name=None, tqdm=False) -> TimeSo
...
         tqdm : bool (default=False)
             Prints tqdm progress bar if set to True.
 
         Returns
         -------
         output : TimeSolverOutput
             Collected values of specified quantities at specified timepoints.
+            If store_data=True, this returns an empty dictionary.
```
ilateur: Do you mean `False`?
Also this feels a bit strange... would returning `None` (so no return) be better? so `if store_data: return output`

mumaxplus/timesolver.py
```diff
@@ -23,6 +26,12 @@ def __init__(self, quantity_dict, file_name=None):
 
         self._keys = list(self._data.keys())  # keep list of keys to maintain order
         self._file_name = file_name
+        self._store_data = store_data
+
+        if self._file_name is None and self._store_data is False:
+            warnings.warn("The timesolver output will not be saved or stored if `file_name` "
+            "is not specified and `store_data` is set to `False`.")
```
ilateur: Maybe suggest to use `run` instead then?

mumaxplus/timesolver.py
```diff
@@ -31,14 +40,21 @@ def __init__(self, quantity_dict, file_name=None):
 
     def write_line(self, time):
         """Compute all the specified quantities for the current state."""
-        self._data["time"].append(time)
+
+        # Compute values
+        values = {"time": time}
         for key, func in self._quantities.items():
-            self._data[key].append(func())
+            values[key] = func()
```
ilateur: I wonder if constantly creating a new dictionary has a noticeable impact on performance, and if it would be better to keep this little `values` dictionary around (`self.`) for all temporary evaluation?
(or does that somewhat defeat the point?)

ilateur (changes requested):

mumaxplus/timesolver.py
lamoreel: The main overhead here comes from the quantity function evaluations and (opening +) writing to a file. Since this PR enables three different ways to use `timesolver.solve()`, this seemed like the best option.

ilateur: I tested it by running the `solve` part of standard problem 4 100 times. On `dev` (before this change), every `solve` takes (2.181 ± 0.014) s. On this branch every solve takes (2.205 ± 0.014) s, but after making `values` a property of TimeSolverOutput it takes (2.1844 ± 0.0095) s again.

This is barely any change, but it *is* noticeable :D 
(Am I allowed to implement this little change?)

ilateur (approved):

PR #135, opened 2026-03-26
Fix Slonczewski STT free layer thickness test

ilateur:
As pointed out by @jplauzie in [PR 391 of mumax3](https://github.com/mumax/3/pull/391), the `free_layer_thickness` test was broken. When fixing this test, I found multiple other small issues that needed fixing as well, such as the `free_layer_thickness` docstring, the differing default value of `Lambda` and non-normalized `fixed_layer`.


mumaxplus/ferromagnet.py
```diff
@@ -436,8 +438,8 @@ def Lambda(self, value):
     
     @property
     def free_layer_thickness(self) -> Parameter:
-        """Slonczewski free layer thickness (m). If set to zero (default),
-        then the thickness will be deduced from the mesh size.
+        """Slonczewski free layer thickness (m). By default, the thickness is
+        deduced from the mesh size.
```
lamoreel: This should probably mention the `z` direction.

src/physics/ferromagnet.cpp
```diff
@@ -34,7 +34,7 @@ Ferromagnet::Ferromagnet(std::shared_ptr<System> system_ptr,
       gamma(system(), 1.7595E11, name + ":gamma", "rad/Ts"),
       temperature(system(), 0.0, name + ":temperature", "K"),
       xi(system(), 0.0, name + ":xi", ""),
-      Lambda(system(), 0.0, name + ":lambda", ""),
+      Lambda(system(), 1.0, name + ":lambda", ""),
```
lamoreel, line 37: This new default value omits the need of `Lambda` inside `SlonczewskiSTTAssuredZero()`.

lamoreel (changes requested):

src/physics/ferromagnet.cpp
ilateur: What if you manually set it to 0 though :thinking:

test/test_free_layer_thickness.py
```diff
@@ -9,13 +9,13 @@
 RTOL = 1e-5  # 0.001%
 
 # Arbitrary parameters, resulting in a non-zero Slonczewski torque
-magnetization = (np.random.uniform(), np.random.uniform(), np.random.uniform())
-magnetization /= np.linalg.norm(magnetization)
+# and arbitrary but consistent magnetization
+magnetization = (0.8096410632133484, -0.29036206007003784, -0.5100698471069336)
```
JonathanMaes: This value seems very random, can't a random magnetization be used that is then copied into the bottom layers of the enlarged grid in the second half of the script? Or would that make for a random test that could fail at random times?

ilateur: Yes it can be random, but yes that was my fear... although this is such a simple test with values that should always be exactly the same so I guess it's okay here?

ilateur: I ran it 10000 times with a random magnetization and it always passed :+1:

lamoreel (approved):

JonathanMaes (approved):

test/test_free_layer_thickness.py
jplauzie: Since mumax+ changes the rng each run, using a random magnetization seems a bit more robust (mumax3 doesn't, which is why I just grabbed the hardcoded value instead of using a m.getcell()). It should always pass, as long as the two magnetizations are the same.

The only time it might be an issue is the 1 in bajillion odds you hit something where torques are ~0/0?  Like if it matches the fixed layer 1,0,1 ~exactly

jplauzie: Speaking of (1,0,1), that should give 0 torque if i set it intentionally, right? I get a nonzero torque value for 
magnetization = (1/np.sqrt(2),0,1/np.sqrt(2)), the torques come out as 
[-0.00019573  0.          0.00019573]
[-0.00019573  0.          0.00019573]

ilateur: Seeing as sqrt(2) can't be perfectly represented as a floating point number, and a typical torque (with these parameters) is something like `[16751.709  -4639.6943 11682.772]`, which is about 1e-8 times smaller, so zero within floating point precision :)

For some reason, when I use `magnetization = (1, 0, 1)` and let mumax normalize itself, I get even closer to zero: `[-5.299274e-06  0.000000e+00  5.299274e-06]`, or ~1e-10 as small :thinking:

jplauzie: Ok, I think that makes sense. I was expecting them to be close to mumax3's values (~1e-7 for m= (0.8096410632133484, -0.29036206007003784, -0.5100698471069336), and ~1e-15 for (1,0,1) ) which is what i was comparing to in my head, since the parameters (msat,jcur,pol,fixedlayer,lambda) seem to be the copied over. I guess something like epsilonprime or alpha must be being initialized differently implicitly somewhere

ilateur: No, the unit of torques is different :) mumax3 divides by gamma, while mumax+ uses the 'pure' `rad / s`

jplauzie: That would do it! Thanks

PR #137, opened 2026-03-30
Add InterParameter tests

ilateur:
I finally added tests for the `InterParameter` class with the help of Claude. Don't worry, it's very heavily curated, but now I'm fairly certain it tests for all important aspects.

### DIFF
--- a/test/test_interparameter.py
+++ b/test/test_interparameter.py
@@ -81,7 +81,7 @@ class TestSetGetBetween:
 
     def test_get_between_after_set_between(self, interex):
         interex.set_between(1, 3, 42.0)
-        assert interex.get_between(1, 3) == pytest.approx(42.0)
+        assert interex.get_between(1, 3) == 42.0
 
     def test_get_between_is_symmetric(self, interex):
         interex.set_between(2, 4, 7.0)
--- a/test/test_interparameter.py
+++ b/test/test_interparameter.py
@@ -1,0 +1,133 @@
+"""Tests for the InterParameter class."""
+
+import numpy as np
+import pytest
+from mumaxplus import World, Grid, Ferromagnet
+
+
+def lut_index(i, j):
+    """Mirror the C++ getLutIndex: lower-triangular LUT index for (i, j)."""
+    if i > j:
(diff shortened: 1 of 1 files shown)


test/test_interparameter.py
```diff
@@ -0,0 +1,133 @@
...
+    """Return a Ferromagnet whose cells are randomly assigned to n_regions."""
+    world = World((1e-9, 1e-9, 1e-9))
+    regions = np.random.randint(0, n_regions, size=(nz, ny, nx))
+    magnet = Ferromagnet(world, Grid((nx, ny, nz)), regions=regions)
+    return magnet
+
+@pytest.fixture
+def interex(magnet):
+    """The inter_exchange InterParameter of the re-used module-scoped magnet
+    fixture. inter_exchange is reset to a known uniform zero state before each
```
lamoreel, line 28: Capitalize.

test/test_interparameter.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+class TestSetGetBetween:
+    def test_set_between_makes_non_uniform(self, interex):
+        interex.set_between(0, 1, 99.0)
+        assert interex.is_uniform is False
+
+    def test_get_between_after_set_between(self, interex):
+        interex.set_between(1, 3, 42.0)
+        assert interex.get_between(1, 3) == pytest.approx(42.0)
```
lamoreel: This should be exactly equal (assuming no rounding errors in simply passing floats/doubles to and from the C++ side). If we want `approx()`, we should also account for double precision.

lamoreel (changes requested):
Great to finally have these tests (without questioning our work ethics...)! ❤️

test/test_interparameter.py
ilateur: Oops, I let one approx slip through

test/test_interparameter.py
ilateur: But `inter_exchange` is a property name? I could add backticks?

lamoreel (approved):

PR #139, opened 2026-03-31
Fix #138: DMI values are taken from origin

### LINKED ISSUE #138: Issue setting DMI value by region
Hi,

I’ve been running into an issue when I try to set the DMI value to be different across multiple regions. During energy calculations and state minimizations, the solver is ignoring the regional assignments. Instead, it seems to be sampling the DMI value from the origin cell and applying it to the entire simulation space.

I verified this by testing a skyrmion on a flat film (4 active spin cells thick), which yields the expected energy results. However, if I pad the simulation space with a layer of empty spin cells below the film (shifting the origin to an empty cell) the DMI energy drops to 0, while anisotropy, demag, and exchange energies remain correct. If I put the empty layer above the sample (keeping the origin inside the active film), I get the correct DMI energy again.

No Empty:
pol Skyrmion r (m)	Skyrmion w (m)	E_ex (J)	              E_DMI (J)	          E_Ku (J)	                    E_demag (J)	              E_tot (J)
+1.0	3.300e-08	1.650e-08	+2.385432e-18	-4.407660e-18	         -3.948414e-17	+1.033616e-17	-3.117021e-17
-1.0	3.300e-08	1.650e-08	+2.385432e-18	+4.407660e-18	-3.948414e-17	         +1.033616e-17	-2.235489e-17

Bottom Layer Empty:
pol	kyrmion r (m)	Skyrmion w (m)	E_ex (J)	          E_DMI (J)	          E_Ku (J)	           E_demag (J)	            E_tot (J)
+1.0	3.300e-08	1.650e-08	+2.385432e-18	+0.000000e+00	-3.948414e-17	+1.033616e-17	-2.676255e-17
-1.0	3.300e-08	1.650e-08	+2.385432e-18	+0.000000e+00	-3.948414e-17	+1.033616e-17	-2.676255e-17

Top Layer Empty
polSkyrmion r (m)	Skyrmion w (m)	E_ex (J)	          E_DMI (J)	         E_Ku (J)	             E_demag (J)	E_tot (J)
+1.0	3.300e-08	1.650e-08	+2.385432e-18	-4.407660e-18	        -3.948414e-17	+1.033616e-17	-3.117021e-17
-1.0	3.300e-08	1.650e-08	+2.385432e-18	+4.407660e-18	-3.948414e-17	+1.033616e-17         -2.235489e-17

[DMI_energy_problems.py](https://github.com/user-attachments/files/26358914/DMI_energy_problems.py)

lamoreel:
The bug is reported and described in #138. The `harmonicMean` member function of `CuParameter` was called recursively, thus casting the real parameter values into `int`s (zeros), returning the value (during the second call) at grid index zero, which is the origin.
This is fixed by bypassing this member function and forcing to look inside the global namespace only.

### DIFF
--- a/src/core/parameter.hpp
+++ b/src/core/parameter.hpp
@@ -106,7 +106,7 @@ __device__ inline real CuParameter::valueAt(int3 coo) const {
 
 __device__ inline real CuParameter::harmonicMean(int idx1, int idx2) const {
   if (idx1 == idx2) { return valueAt(idx1); }
-  else { return harmonicMean(valueAt(idx1), valueAt(idx2)); }
+  else { return ::harmonicMean(valueAt(idx1), valueAt(idx2)); }
 }
 
 


lamoreel:
The bug does not occur when there's no geometry or regions, because in that case `uniformValue_` is returned regardless of the index of the cell in question.

ilateur (approved):
I was also thinking that it was a recursive issue, so I approve :)

PR #141, opened 2026-04-10
Altermagnets

lamoreel:
This pull request adds support for simulating d-wave altermagnets through a new `Altermagnet` class. The main difference with `Antiferromagnet` is the removal of the ferromagnetic exchange field in favor of an anisotropic field. In principle, the user is still able to take both fields into account, but that's their responsibility :blush:.

For the time being, only open boundaries and 2D magnets are supported in the anisotropic exchange.


test/test_full_magnetization.py
```diff
@@ -23,6 +23,22 @@ def test_full_magnetization_afm():
     err = max_relative_error(result, wanted)
     assert err < 1e-5  # only error should be single versus double precision
 
+def test_full_magnetization_atm():
```
DiegoDeGusem, line 26: This test passes, but gives the warning `The anisotropic exchange interaction in altermagnets is only supported for two-dimensional grids in the xy-plane. This interaction will have no effect in the z-direction.` It feels a bit strange to have a warning message in tests.

examples/spinwave_dispersion_atm.py
```diff
@@ -0,0 +1,155 @@
...
+
+    # Run solver
+    m1 = np.zeros(shape=(nt, 3, 1, 1, nx))
+    m2 = np.zeros(shape=(nt, 3, 1, 1, nx))
+    m1[0,...] = magnet.sub1.magnetization.eval()
+    m2[0,...] = magnet.sub2.magnetization.eval()
+    for i in tqdm(range(nt)):
+        world.timesolver.run(dt)
+        m1[i,...] = magnet.sub1.magnetization.eval()
+        m2[i,...] = magnet.sub2.magnetization.eval()
```
ilateur: index 0 is overwritten again

examples/spinwave_dispersion_atm.py
```diff
@@ -0,0 +1,155 @@
...
+y_end   = int((ymax - extent[2]) / (extent[3] - extent[2]) * nt)
+
+m1_max = np.max(m1_FT[y_start:y_end, x_start:x_end])
+m2_max = np.max(m2_FT[y_start:y_end, x_start:x_end])
+
+FT_tot1 = m1_FT/m1_max
+FT_tot2 = m2_FT/m2_max
+
+# Sum of sublattice spectra produces the nicest plot
+FT_tot = FT_tot1 + FT_tot2
```
ilateur, line 115: You can sum images instead of data to use different cmaps for each if you want :)
Or don't, this is perfectly fine

src/physics/mumaxworld.hpp
```diff
@@ -89,9 +101,58 @@ class MumaxWorld : public World {
...
 
+  // ----------------------------------------------------------------------------------
+  // -------------------------------- Helper functions --------------------------------
+  // ----------------------------------------------------------------------------------
 
-  // --------------------------------------------------
-  // PBC
+  // * Helper function to add any magnet to the world
+  template <class T, class MapT>
+  T* addMagnetTempl(MapT& container,
```
ilateur: I would've expected `std::map<std::string, std::unique_pointer<T>>` instead of `template <class MapT>` 🤔

src/physics/quantityevaluator.hpp
```diff
@@ -102,6 +104,7 @@ class ScalarQuantityEvaluator : public ScalarQuantity {
 };
 
 using AFM_ScalarQuantity = ScalarQuantityEvaluator<Antiferromagnet>;
+using ATM_ScalarQuantity = ScalarQuantityEvaluator<Altermagnet>;
 using FM_ScalarQuantity = ScalarQuantityEvaluator<Ferromagnet>;
 using M_ScalarQuantity = ScalarQuantityEvaluator<Magnet>;
 using NcAfm_ScalarQuantity = ScalarQuantityEvaluator<NcAfm>;
```
ilateur, line 110: This is getting a bit silly, is it still the best way to do things? 🤔 I guess so?

src/bindings/wrap_altermagnet.cpp
```diff
@@ -0,0 +1,57 @@
...
+void wrap_altermagnet(py::module& m) {
+  py::class_<Altermagnet, Magnet>(m, "Altermagnet")
+      .def("sub1", &Altermagnet::sub1, py::return_value_policy::reference)
+      .def("sub2", &Altermagnet::sub2, py::return_value_policy::reference)
+      .def("sublattices", &Altermagnet::sublattices, py::return_value_policy::reference)
+      .def("other_sublattice",
+          [](const Altermagnet* m, Ferromagnet* mag) { return m->getOtherSublattices(mag)[0]; },
+            py::return_value_policy::reference)
+      .def_readonly("atmex_cell", &Altermagnet::afmex_cell)
+      .def_readonly("atmex_nn", &Altermagnet::afmex_nn)
```
ilateur: I still don't see why these should be renamed while doing exactly the same. Even in an Altermagnet they still have an antiferromagnetic contribution, right?

test/test_domain_wall_atm.py
```diff
@@ -0,0 +1,124 @@
...
+
+SRTOL = 5e-3
+
+cs = 0.5e-9 # Small to get good resolution in net magnetization profile
+nx, ny, nz = 512, 1, 1
+dw = 14  # Initial guess of number of cells in DW
+
+def max_relative_error(result, wanted):
+    err = np.abs(result - wanted)
+    relerr = err / np.max(np.abs(wanted))
```
ilateur, line 23: Not a true relative error with this denominator

src/physics/altermagnet.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include "relaxer.hpp"
+
+Altermagnet::Altermagnet(std::shared_ptr<System> system_ptr,
+                         std::string name)
+    : HostMagnet(system_ptr, name),
+      sub1_(Ferromagnet(system_ptr, name + ":sublattice_1", this)),
+      sub2_(Ferromagnet(system_ptr, name + ":sublattice_2", this)),
+      alterex_1(system(), 0.0, name + ":alterex_1", "J/m"),
+      alterex_2(system(), 0.0, name + ":alterex_2", "J/m"),
+      alterex_angle(system(), 0.0, name + ":alterex_angle", ""),
```
ilateur: `"rad"` as unit? We did that with the angleFieldQuantity.

mumaxplus/altermagnet.py
```diff
@@ -0,0 +1,572 @@
+"""Altermagnet implementation."""
```
ilateur, line 1: Half of this (fairly large) file is copy-pasted and unchanged from the Antiferromagnet class :(

DiegoDeGusem (changes requested):
Nice and very clean! Just one general comment. The altermagnets are not included on the website. They should be added in `api.rst`.

src/physics/atmexchange.cu
```diff
@@ -0,0 +1,227 @@
...
+#include "atmexchange.hpp"
+#include "ferromagnet.hpp"
+#include "field.hpp"
+#include "inter_parameter.hpp"
+#include "parameter.hpp"
+#include "reduce.hpp"
+#include "world.hpp"
+
+bool atmExchangeAssuredZero(const Ferromagnet* magnet) {
+  if (!magnet->hostMagnet()->asATM()) { return true; }
```
ilateur, line 16: I suppose this is also safe if there is no hostMagnet at all?

src/physics/atmexchange.cu
```diff
@@ -0,0 +1,227 @@
...
+#pragma unroll
+  for (int i = 0; i < 4; ++i) {
+    coos[i] = mastergrid.wrap(coo + rel_coos[i]);
+    inBulk &= hField.cellInGeometry(coos[i]);
+    inBulk &= (msat.valueAt(coos[i]) != 0);
+  }
+
+  if (!inBulk) {
+    hField.setVectorInCell(idx, h / msat.valueAt(idx));
+    return;
```
ilateur, line 134: So at any edge/corner there simply is no mixed derivative at all?

ilateur: I guess this is a type of open boundary condition?

ilateur (changes requested):
I would still like to leverage the object-oriented approach by making the `Altermagnet` a child class of the `Antiferromagnet`. It has exactly the same functionality, but expanded. Even if physically these are totally different materials, the implementation of shared aspects is identical. Though I must admit, the amout of copy-pasting and function-tweaking is less bad than I expected.

Other than that, everything looks clean to me with some very nice tests.

ilateur (commented):
aaa

ilateur (changes requested):
why do you delete

test/test_domain_wall_atm.py
ilateur, line 1: Could this test be made extra robust by also making a magnet in the y-direction and setting `alterex_angle` to `np.pi/2`? (or something like that)

src/physics/atmexchange.cu
lamoreel: A very crude one, yes.

src/bindings/wrap_altermagnet.cpp
lamoreel: Okay, you've convinced me.

src/physics/quantityevaluator.hpp
lamoreel: There is probably a cleaner way of doing things, but this may lie outside the scope of this PR (since the answer will be non-trivial). Feel free to add an item on the TODO heap!

src/physics/mumaxworld.hpp
lamoreel: I wanted to make it more generic, but I see now that the body always implicitly assumes the mapped type is `std::unique_ptr<T>` so you are very right to question this!

src/physics/quantityevaluator.hpp
ilateur: True, but I also don't see a better way as of right now

mumaxplus/altermagnet.py
```diff
@@ -234,7 +234,7 @@ def alterex_1(self) -> Parameter:
 
         See Also
         --------
-        alterex_2, alterex_angle, atmex_cell, atmex_nn
+        alterex_2, alterex_angle, afmex_cell, afmex_nn
```
ilateur: I was wondering why the afm exchange constants are referred to inside the altermagnetic exchanges

examples/Bloch_wall_altermagnet.py
ilateur, line 1: I like this example quite a lot. Would it be useful to put this on the documentation website too? This can serve as a real-space example of both antiferromagnetic and altermagnetic exchanges. If you don't have the time, but like the idea, I (or Diego) could do it too.

ilateur (commented):
Make sure to resolve the merge conflicts as well

ilateur (approved):

DiegoDeGusem (approved):

PR #142, opened 2026-04-10
Add ovf functions and tests

DiegoDeGusem:
This PR adds `save_ovf` and `load_ovf` methods to `FieldQuantities`. This allows users to save and load `FieldQuantities` via OVF files.
The `save_ovf` method uses `pyovf` and provides it with all available metadata. However, some metadata appears to be ignored (e.g. simulation time) and some fields are hard-coded in pyovf (e.g. units).
`pyovf` expects arrays with shape `(nz, ny, nx, ncomp)`, while mumax⁺ uses `(ncomp, nz, ny, nx)`. The save function reorders the axis, while the load function reverses this. To notify users of this reordering, the save function has a warning in the docstring.

### DIFF
--- a/mumaxplus/fieldquantity.py
+++ b/mumaxplus/fieldquantity.py
@@ -3,6 +3,7 @@
 import numpy as _np
 import pyovf
 from .grid import Grid
+from pathlib import Path
 
 
 class FieldQuantity:
@@ -156,11 +157,12 @@ class FieldQuantity:
         name : str (default="")
             The name of the OVF file. If the name is empty (the default), it will look for the most recently saved file with this FieldQuantity name."""
         if name == "":
-            if self.name not in self._ovf_counts:
-                raise FileNotFoundError(f"{self.name.replace(":", "_") + f"{0:06d}.ovf"} does not exist")
-            count = self._ovf_counts[self.name] - 1
-            name = self.name.replace(":", "_") + f"{count:06d}.ovf"
-
+            name_replace = self.name.replace(":", "_")
+            files = Path(".").glob(f"{name_replace}*.ovf")
+            try:
+                name = str(max(files, key=lambda f: int(f.stem[-6:])))
+            except ValueError:
+                raise FileNotFoundError(F"No files matching '{name_replace}*.ovf' found.")
         ovf = pyovf.read(name)
         if self.ncomp == 1:
             data = _np.array([ovf.data])
--- a/mumaxplus/fieldquantity.py
+++ b/mumaxplus/fieldquantity.py
@@ -1,7 +1,7 @@
 """FieldQuantity implementation."""
 
 import numpy as _np
-
+import pyovf
 from .grid import Grid
 
 
@@ -122,3 +122,34 @@ class FieldQuantity:
             self._impl.exec()
         stop = time.time()
         return (stop - start) / ntimes
+
+    def save_ovf(self, name=""):
+        """Save the FieldQuantity as an OVF file.
+
+        Parameters
+        ----------
+        name : str (default="")
+            The name of the OVF file. If the name is empty (the default), the name of the FieldQuantity will be used.
+            
+        Warning
+        -------
+        The shape of the array in the OVF file is (nz, ny, nx, ncomp) and not (ncomp, nz, ny, nx) in order to have the correct metadata."""
+        cx, cy, cz = self._impl.system.cellsize
+        ovf = pyovf.create(_np.moveaxis(self.eval(), 0, -1), xstepsize=cx, ystepsize=cy, zstepsize=cz, title=self.name)
+        ovf.TotalSimTime = self._impl.system.time
+        if name == "":
+            name = self.name + ".ovf"
+        pyovf.write(name, ovf)
+
+    def load_ovf(self, name=""):
+        """Load an OVF file as a FieldQuantity.
+
+        Parameters
+        ----------
+        name : str (default="")
+            The name of the OVF file. If the name is empty (the default), the name of the FieldQuantity will be used."""
+        if name == "":
+            name = self.name + ".ovf"
+        ovf = pyovf.read(name)
+        data = _np.ascontiguousarray(_np.moveaxis(ovf.data, -1, 0))
+        self.set(data)
--- a/mumaxplus/fieldquantity.py
+++ b/mumaxplus/fieldquantity.py
@@ -151,5 +151,8 @@ class FieldQuantity:
         if name == "":
             name = self.name + ".ovf"
         ovf = pyovf.read(name)
-        data = _np.ascontiguousarray(_np.moveaxis(ovf.data, -1, 0))
+        if self.ncomp == 1:
+            data = _np.array([ovf.data])
+        else:
+            data = _np.ascontiguousarray(_np.moveaxis(ovf.data, -1, 0))
         self.set(data)
--- a/mumaxplus/fieldquantity.py
+++ b/mumaxplus/fieldquantity.py
@@ -145,7 +145,7 @@ class FieldQuantity:
         ovf.TotalSimTime = self._impl.system.time
         if name == "":
             count = self._ovf_counts.get(self.name, 0)
-            name = self.name.replace(":", "_") + f"{count:06d}.ovf"
+            name = self.name.replace(":", "_") + f"_{count:06d}.ovf"
             self._ovf_counts[self.name] = count + 1
         pyovf.write(name, ovf)
 
@@ -160,9 +160,9 @@ class FieldQuantity:
             name_replace = self.name.replace(":", "_")
             files = Path(".").glob(f"{name_replace}*.ovf")
             try:
-                name = str(max(files, key=lambda f: int(f.stem[-6:])))
+                name = str(max(filter(lambda f: f.stem[-6:].isnumeric(), files), key=lambda f: int(f.stem[-6:])))
             except ValueError:
-                raise FileNotFoundError(F"No files matching '{name_replace}*.ovf' found.")
+                raise FileNotFoundError(F"No files matching '_{name_replace}*.ovf' found.")
         ovf = pyovf.read(name)
         if self.ncomp == 1:
             data = _np.array([ovf.data])
--- a/mumaxplus/fieldquantity.py
+++ b/mumaxplus/fieldquantity.py
@@ -7,6 +7,7 @@ from .grid import Grid
 
 class FieldQuantity:
     """A functor representing a physical field quantity."""
+    _ovf_counts = {} # It keeps resetting to 0 if it is in the __init__
     def __init__(self, impl):
         self._impl = impl
 
@@ -133,23 +134,27 @@ class FieldQuantity:
             
         Warning
         -------
-        The shape of the array in the OVF file is (nz, ny, nx, ncomp) and not (ncomp, nz, ny, nx) in order to have the correct metadata."""
+        The shape of the array in the OVF file is (nz, ny, nx, ncomp) and not (ncomp, nz, ny, nx) in order to have the correct metadata.
+        
+        Warning
+        -------
+        self.name returns a string with colons (:). To avoid issues on Windows, these colons are changed to underscores (_)."""
         cx, cy, cz = self._impl.system.cellsize
         ovf = pyovf.create(_np.moveaxis(self.eval(), 0, -1), xstepsize=cx, ystepsize=cy, zstepsize=cz, title=self.name)
         ovf.TotalSimTime = self._impl.system.time
         if name == "":
-            name = self.name.replace(":", "_") + ".ovf"
+            count = self._ovf_counts.get(self.name, 0)
+            name = self.name.replace(":", "_") + f"{count:06d}.ovf"
+            self._ovf_counts[self.name] = count + 1
         pyovf.write(name, ovf)
 
-    def load_ovf(self, name=""):
+    def load_ovf(self, name):
         """Load an OVF file as a FieldQuantity.
 
         Parameters
         ----------
         name : str (default="")
-            The name of the OVF file. If the name is empty (the default), the name of the FieldQuantity will be used."""
-        if name == "":
-            name = self.name.replace(":", "_") + ".ovf"
+            The name of the OVF file."""
         ovf = pyovf.read(name)
         if self.ncomp == 1:
             data = _np.array([ovf.data])
--- a/mumaxplus/fieldquantity.py
+++ b/mumaxplus/fieldquantity.py
@@ -138,7 +138,7 @@ class FieldQuantity:
         ovf = pyovf.create(_np.moveaxis(self.eval(), 0, -1), xstepsize=cx, ystepsize=cy, zstepsize=cz, title=self.name)
         ovf.TotalSimTime = self._impl.system.time
         if name == "":
-            name = self.name + ".ovf"
+            name = self.name.replace(":", "_") + ".ovf"
         pyovf.write(name, ovf)
 
     def load_ovf(self, name=""):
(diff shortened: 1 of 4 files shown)


mumaxplus/fieldquantity.py
```diff
@@ -122,3 +122,34 @@ def _bench(self, ntimes=100):
...
+            The name of the OVF file. If the name is empty (the default), the name of the FieldQuantity will be used.
+            
+        Warning
+        -------
+        The shape of the array in the OVF file is (nz, ny, nx, ncomp) and not (ncomp, nz, ny, nx) in order to have the correct metadata."""
+        cx, cy, cz = self._impl.system.cellsize
+        ovf = pyovf.create(_np.moveaxis(self.eval(), 0, -1), xstepsize=cx, ystepsize=cy, zstepsize=cz, title=self.name)
+        ovf.TotalSimTime = self._impl.system.time
+        if name == "":
+            name = self.name + ".ovf"
```
JonathanMaes: `self.name` can contain a colon `:`. On Windows, a colon in a path has a special meaning, indicating a [file "stream"](https://stackoverflow.com/a/25896961).

In `test_ovf.py`, for instance, the file name generated by this line of code would be `ferromagnet_1:magnetization.ovf`. On Windows, the file `ferromagnet_1` (no extension) gets created with the associated stream `magnetization.ovf`. This does not impede the working of `load_ovf()`, which automatically grabs the correct stream, and it is even possible to store multiple streams in one file like this (e.g., both `magnetization` and `bias_magnetic_field`).

However, other programs (e.g., [mumax-view](https://mumax.ugent.be/mumax-view/)) won't expect this behavior, and neither will users, who will only see the empty file `ferromagnet_1` in their filesystem. Therefore, I would simply replace the colon by some other character that has no special meaning.

mumaxplus/fieldquantity.py
```diff
@@ -122,3 +122,34 @@ def _bench(self, ntimes=100):
...
+
+    def load_ovf(self, name=""):
+        """Load an OVF file as a FieldQuantity.
+
+        Parameters
+        ----------
+        name : str (default="")
+            The name of the OVF file. If the name is empty (the default), the name of the FieldQuantity will be used."""
+        if name == "":
+            name = self.name + ".ovf"
```
JonathanMaes: Whatever change occurs to address the colon-issue in `save_ovf` should also be applied here.

mumaxplus/fieldquantity.py
```diff
@@ -122,3 +122,34 @@ def _bench(self, ntimes=100):
...
+        Parameters
+        ----------
+        name : str (default="")
+            The name of the OVF file. If the name is empty (the default), the name of the FieldQuantity will be used.
+            
+        Warning
+        -------
+        The shape of the array in the OVF file is (nz, ny, nx, ncomp) and not (ncomp, nz, ny, nx) in order to have the correct metadata."""
+        cx, cy, cz = self._impl.system.cellsize
+        ovf = pyovf.create(_np.moveaxis(self.eval(), 0, -1), xstepsize=cx, ystepsize=cy, zstepsize=cz, title=self.name)
```
JonathanMaes, line 144: Is this necessary for parameters with 1 component (e.g., `aex`)? It seems they are stored as 3D arrays, so moveaxis shuffles the XYZ-order around.

test/test_ovf.py
```diff
@@ -0,0 +1,56 @@
...
+        self.magnet.msat = 800e3
+        self.magnet.aex = 13e-12
+        self.magnet.alpha = 0.02
+
+        self.world.timesolver.run(0.5e-9)
+
+        self.magnetization = self.magnet.magnetization.eval()
+
+        self.magnet.magnetization.save_ovf(self.name)
+        self.magnet.magnetization.save_ovf()
```
JonathanMaes, line 33: I would add an example of a 1-component field to this test as well (and any other kinds of saveable things I'm missing).

mumaxplus/fieldquantity.py
```diff
@@ -122,3 +122,34 @@ def _bench(self, ntimes=100):
...
         stop = time.time()
         return (stop - start) / ntimes
+
+    def save_ovf(self, name=""):
+        """Save the FieldQuantity as an OVF file.
+
+        Parameters
+        ----------
+        name : str (default="")
+            The name of the OVF file. If the name is empty (the default), the name of the FieldQuantity will be used.
```
ilateur: In mumax3 there is a very nice feature where by default it will take the name of the quantity and append an integer of 6 digits, which is always increasing, so `m000000.ovf`, `m000001.ovf` and so on. This could be nice to have here as well?

Although I don't know how easy this number is to keep track of inside the python layer, where a new python `FieldQuantity` instance is created every time and wrapped around the same `_impl` instance.

mumaxplus/fieldquantity.py
```diff
@@ -122,3 +122,34 @@ def _bench(self, ntimes=100):
...
+            name = self.name + ".ovf"
+        pyovf.write(name, ovf)
+
+    def load_ovf(self, name=""):
+        """Load an OVF file as a FieldQuantity.
+
+        Parameters
+        ----------
+        name : str (default="")
+            The name of the OVF file. If the name is empty (the default), the name of the FieldQuantity will be used."""
```
ilateur: Should this docstring have a warning too?

mumaxplus/fieldquantity.py
```diff
@@ -122,3 +122,34 @@ def _bench(self, ntimes=100):
...
+        if name == "":
+            name = self.name + ".ovf"
+        ovf = pyovf.read(name)
+        data = _np.ascontiguousarray(_np.moveaxis(ovf.data, -1, 0))
```
ilateur: maybe add a lil comment

src/bindings/wrap_system.cpp
```diff
@@ -40,6 +40,9 @@ py::array_t<T> get_data(const System* system, Getter getter, T default_value) {
 void wrap_system(py::module& m) {
   py::class_<System, std::shared_ptr<System>>(m, "System")
       .def_property_readonly("grid", &System::grid)
+      .def_property_readonly("time", [](const System& s) {
+                                     return s.world()->time();
+                                     })
```
ilateur: (I'd make this 1 line)

test/test_ovf.py
```diff
@@ -0,0 +1,56 @@
+from mumaxplus import *
+import pyovf
+import numpy as np
+from mumaxplus.util import plot_field
```
ilateur: unused

test/test_ovf.py
```diff
@@ -0,0 +1,56 @@
+from mumaxplus import *
```
ilateur: It's a bit better to only import what is needed

test/test_ovf.py
```diff
@@ -0,0 +1,56 @@
...
+        self.world = World(cellsize=(cx, cy, cz))
+
+        self.magnet = Ferromagnet(self.world, Grid((nx, ny, nz)))
+
+        self.magnet.magnetization = (1,0,0)
+        self.magnet.msat = 800e3
+        self.magnet.aex = 13e-12
+        self.magnet.alpha = 0.02
+
+        self.world.timesolver.run(0.5e-9)
```
ilateur: Why set these parameters and why set uniform and run? Isn't a random magnetization fine?

test/test_ovf.py
```diff
@@ -0,0 +1,56 @@
...
+
+        self.magnet.magnetization.save_ovf(self.name)
+        self.magnet.magnetization.save_ovf()
+
+        self.ovf = pyovf.read(self.name)
+
+    def test_read(self):
+        self.magnet.magnetization.load_ovf()
+        mag = self.magnet.magnetization.eval()
+        assert max_relative_error(mag, self.magnetization) < RTOL
```
ilateur: This test is not that robust. If `load_ovf` does nothing at all, it would still succeed. Maybe change the magnetization to something different before the load function?

test/test_ovf.py
```diff
@@ -0,0 +1,56 @@
...
+RTOL = 2e-7
+
+def max_relative_error(result, wanted):
+    return np.max(np.abs(result - wanted) / np.abs(wanted))
+
+class TestOVF:
+    def setup_class(self):
+        self.name = "test.ovf"
+        cx, cy, cz = 4e-9, 2e-9, 5e-9
+        nx, ny, nz = 128, 32, 1
```
ilateur: A smaller (but 3D) grid would be fine for this basic test I think

test/test_ovf.py
ilateur, line 1: Maybe this test should also try writing/reading a `Parameter` with 1 component (like `aex`) or a raw `FieldQuantity` with 6 components like `conductivity_tensor`?

ilateur (changes requested):

JonathanMaes (changes requested):

mumaxplus/fieldquantity.py
ilateur: These are also stored as 4D arrays, just with (1, cz, cy, cx), at least in mumax+

mumaxplus/fieldquantity.py
DiegoDeGusem: I didn't do that because it's not that relevant for the user

test/test_ovf.py
DiegoDeGusem: Random would work but takes longer to run. The reason I wanted to run was to check if the time was saved in the ovf. But even though I give the time to pyovf, it doesn't place it in the ovf file

mumaxplus/fieldquantity.py
DiegoDeGusem: I changed it to underscores

JonathanMaes (approved):

test/test_ovf.py
ilateur: You also never check for the time in this test

mumaxplus/fieldquantity.py
```diff
@@ -133,23 +134,27 @@ def save_ovf(self, name=""):
...
         ovf.TotalSimTime = self._impl.system.time
         if name == "":
-            name = self.name.replace(":", "_") + ".ovf"
+            count = self._ovf_counts.get(self.name, 0)
+            name = self.name.replace(":", "_") + f"{count:06d}.ovf"
+            self._ovf_counts[self.name] = count + 1
         pyovf.write(name, ovf)
 
-    def load_ovf(self, name=""):
+    def load_ovf(self, name):
```
JonathanMaes: Is there some fundamental limitation why the most recently saved file (as defined by `_ovf_counts`) can't be loaded?

mumaxplus/fieldquantity.py
```diff
@@ -148,13 +148,19 @@ def save_ovf(self, name=""):
...
         """Load an OVF file as a FieldQuantity.
 
         Parameters
         ----------
         name : str (default="")
-            The name of the OVF file."""
+            The name of the OVF file. If the name is empty (the default), it will look for the most recently saved file with this FieldQuantity name."""
+        if name == "":
+            if self.name not in self._ovf_counts:
+                raise FileNotFoundError(f"{self.name.replace(":", "_") + f"{0:06d}.ovf"} does not exist")
```
JonathanMaes: I believe this is not entirely true. If the script is run a second time, `self._ovf_counts` will be empty, but the file exists nonetheless.
Another possible improvement, which is probably too hand-holdy, could be to enumerate all the files associated with this `FieldQuantity` and load the one with the highest number, throwing an error if no file is found.

mumaxplus/fieldquantity.py
```diff
@@ -156,11 +157,12 @@ def load_ovf(self, name=""):
...
         if name == "":
-            if self.name not in self._ovf_counts:
-                raise FileNotFoundError(f"{self.name.replace(":", "_") + f"{0:06d}.ovf"} does not exist")
-            count = self._ovf_counts[self.name] - 1
-            name = self.name.replace(":", "_") + f"{count:06d}.ovf"
-
+            name_replace = self.name.replace(":", "_")
+            files = Path(".").glob(f"{name_replace}*.ovf")
+            try:
+                name = str(max(files, key=lambda f: int(f.stem[-6:])))
```
JonathanMaes: This throws an error when at least 1 file whose name starts with `name_replace` but doesn't have 6 digits at the end exists in the current directory. This can be alleviated by, e.g.,
```python
name = str(max(filter(lambda f: f.stem[-6:].isnumeric(), files), key=lambda f: int(f.stem[-6:])))
```

mumaxplus/fieldquantity.py
```diff
@@ -129,16 +131,22 @@ def save_ovf(self, name=""):
...
+        -------
+        self.name returns a string with colons (:). To avoid issues on Windows, these colons are changed to underscores (_)."""
         cx, cy, cz = self._impl.system.cellsize
         ovf = pyovf.create(_np.moveaxis(self.eval(), 0, -1), xstepsize=cx, ystepsize=cy, zstepsize=cz, title=self.name)
         ovf.TotalSimTime = self._impl.system.time
         if name == "":
-            name = self.name + ".ovf"
+            count = self._ovf_counts.get(self.name, 0)
+            name = self.name.replace(":", "_") + f"{count:06d}.ovf"
+            self._ovf_counts[self.name] = count + 1
```
ilateur, line 149: that's pretty clever

Maybe also add an underscore between name and count (`+ f"_{count:06d}.ovf"`)? This is not the mumax3 way though, so what do you think?

mumaxplus/fieldquantity.py
```diff
@@ -129,16 +131,22 @@ def save_ovf(self, name=""):
         Parameters
         ----------
         name : str (default="")
-            The name of the OVF file. If the name is empty (the default), the name of the FieldQuantity will be used.
+            The name of the OVF file. If the name is empty (the default), the name of the FieldQuantity will be used appended with an integer of 6 digits.
```
ilateur, line 134: Maybe if the name is not empty, it should be checked whether it ends with ".ovf" and append it if needed?

mumaxplus/fieldquantity.py
```diff
@@ -147,9 +155,18 @@ def load_ovf(self, name=""):
         Parameters
         ----------
         name : str (default="")
-            The name of the OVF file. If the name is empty (the default), the name of the FieldQuantity will be used."""
+            The name of the OVF file. If the name is empty (the default), it will look for the most recently saved file with this FieldQuantity name."""
         if name == "":
-            name = self.name + ".ovf"
+            name_replace = self.name.replace(":", "_")
+            files = Path(".").glob(f"{name_replace}*.ovf")
```
ilateur, line 161: I'm worried this might be a bit slow when you have saved a lot of files during your simulation. Also having a default behavior for load at all is a bit strange to me 🤔 I would take the simple option and just make `name` a necessary variable.

ilateur (commented):
It's a bit funny and sad that we now have nice functionality to save and load ovf files... but not npy files. Would it be nice to also have a `save_npy()` function? Or make it one `save()` function and decide the type based on the given string (or default if not given)? Or is this useless (just call `np.save(field_quantity.eval())`) or outside the scope of this pull request?

mumaxplus/fieldquantity.py
JonathanMaes: If a situation arises where `glob` somehow becomes a performance concern, the user can always specify `name` explicitly, thereby avoiding this entire block of code.

mumaxplus/fieldquantity.py
DiegoDeGusem: I think it is clearer with underscores, but I wanted to follow mumax³

ilateur: The names are already completely different, so breaking this too is not that bad I think

JonathanMaes (approved):

ilateur (approved):

PR #145, opened 2026-04-23
Fix warning message formatting for FP_PRECISION

Romeo75:
Updated warning message to use single quotes for environment variable.


Romeo75:
I just installed the software. However, during the execution of the test script "python examples/standardproblem4.py --mumaxplus-fp-precision DOUBLE", we detected an f-string error at the modified line. We propose this solution as it allowed us to run the test script.

JonathanMaes (approved):
Both the previous and the new version work fine for me (Windows, Python 3.13).

PR #147, opened 2026-04-24
C++ Class Diagrams

DiegoDeGusem:
This PR would automatically generate class diagrams each time the website gets build and they will be added to the website. There are a total of 6 diagrams to avoid a cluttered image.
### Class Diagrams
**Parameters**
image
**Magnetics**
image
**Strayfield**
image
**Time Integration**
image
**Domain**
image
**Utilities**
image


docs/conf.py
```diff
@@ -125,4 +130,12 @@
     ],
     "icon_links_label": "Quick Links",  # screen reader label
     "use_download_button": False,
-}
\ No newline at end of file
+}
+
+def setup(app):
+    project_root = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+    subprocess.run(['clang-uml'], cwd=project_root)
```
JonathanMaes: Does the user have to install `clang-uml` manually?

docs/conf.py
```diff
@@ -33,7 +35,8 @@
     "sphinxcontrib.video",
     "sphinx_copybutton",
     "sphinx.ext.napoleon",
-    "sphinx_design"
+    "sphinx_design",
+    "sphinxcontrib.plantuml"
```
JonathanMaes, line 41: This must be added to the `environment.yml` as well (all the way at the end, as `  - sphinxcontrib-plantuml`).

While you are editing `environment.yml`, another unrelated warning I get is `FutureWarning: Adding 'defaults' to channel list implicitly is deprecated and will be removed in 25.9.`, which should be resolvable by adding `  - defaults` to the `channels:` since conda is already at 26.3.2 nowadays.

JonathanMaes (changes requested):

environment.yml
```diff
@@ -40,3 +42,6 @@ dependencies:
   - sphinxcontrib-video
   - sphinx-copybutton
   - sphinx-design
+  - clang-uml
```
JonathanMaes: This package is [only available](https://anaconda.org/channels/bkryza/packages/clang-uml/overview) on linux, so this sadly makes the environment creation throw an error on Windows :/

Perhaps a better option is to just add a note to the documentation section of the README that [clang-uml](https://github.com/bkryza/clang-uml/releases) must be installed manually (and be in the PATH) in order to generate the diagrams. A descriptive warning could be added in the `make html` if `clang-uml` is not detected.

However, while trying to get this to work on Windows, I keep getting new errors whenever I resolve one. Since these diagrams probably only have to be generated by us for the website, I think it should suffice to put the `clang-uml` command invocation in an `if` block that only runs on linux, and warn that the diagrams are not generated on Windows.

JonathanMaes (changes requested):

environment.yml
DiegoDeGusem: Would you then still gitignore the generated diagrams? If so, then the `make html` command would give warnings that it doesn't find the .puml files.

docs/conf.py
```diff
@@ -141,6 +141,19 @@ def setup(app):
     ], check=True, cwd=project_root)
     subprocess.run(['clang-uml'], cwd=project_root)
 
+conf_dir = os.path.dirname(os.path.abspath(__file__))
+clang_uml_path = os.path.join(conf_dir, "../.clang-uml")
+
+print(clang_uml_path)
```
ilateur: heh?

environment.yml
JonathanMaes: I think it is fine to have those warnings, then if anyone tries to build it on Windows regardless, they will still be able to generate most of the website but they will see clear warnings that something is off.

JonathanMaes (approved):

docs/diagrams/plantuml.jar
ilateur, line 1: Damn, big file

.gitignore
ilateur, line 1: Just checking, these are generated files, but are themselves not needed by the website?

README.md
```diff
@@ -145,7 +145,7 @@ It follows the [NumPy style guide](https://numpydoc.readthedocs.io/en/latest/for
 make html
 ```
 
-The documentation can then be found at `docs/_build/html/index.html`.
+The documentation can then be found at `docs/_build/html/index.html`. However, to generate the class diagrams you should be on a Linux device and first install [clang-uml](https://github.com/bkryza/clang-uml/tree/master) otherwise they will not be generated.
```
ilateur, line 37: Why not Windows?

docs/conf.py
```diff
@@ -125,4 +132,42 @@
...
+        warnings.warn(
+            "UML diagram generation is not supported on Windows. Skipping.",
+            UserWarning
+        )
+        return
+    
+    if shutil.which("clang-uml") is None:
+        warnings.warn(
+            "clang-uml not found. Skipping UML diagram generation. "
+            "Install it from https://github.com/bkryza/clang-uml or via your package manager.",
```
ilateur: this feels as a bit of a strange warning now that it should be inside the environment 🤔

ilateur (changes requested):
I was unable to generate these images myself (on monster5). I tried installing clang-uml via conda but I got errors with `make html` (I did not try to solve them). I am not allowed to install clang-uml via the package manager there, but Lars tried and failed that install.

It would be best if everything was included inside the environment.yml as a self-contained package that works for everybody, but somehow that did not work on Windows?

The actual diagrams look nice, although I wasn't able to judge them on the website. Sadly I don't think we should merge this if only 1 person is able to successfully run it.

DiegoDeGusem:
I will move this back to draft and see if I can fix the issues.

ilateur:
It was successfully built on monster5!

Though, I must admit it's hard to keep an overview when all boxes are that large. Either you see the connections, but can't read anything, or you zoom in to read, but can only look at a single class.

DiegoDeGusem:
### Class Diagrams second try
Unfortunately the Conda package `clang-uml` only works on Linux when you install it via Conda, so I removed it from the environment. You are still able to generate the website without the package, but then you get non-fatal errors, because the diagrams won't be generated. I also added a warning when you generate the documentation that should guide users to the installation of `clang-uml` if they want to (we need it when we want to regenerate the website when pushing to dev).

I was able to generate the diagrams (and the rest of the documentation) on my Linux device and monster5 using both the **Conda** method and the **Ubuntu** method. I don't know if it now works on Windows if you install it via the [clang-uml](https://clang-uml.github.io/md_docs_2installation.html) installation guide (I think it _should_ work :crossed_fingers:) @JonathanMaes could you check this?

I have also changed the layout of the diagrams a little bit to be more like the previous .drawio diagram. I removed the constructor and deconstructor and I also changed the layout from left to right to from top to bottom, which feels cleaner.

**Quantities**
image
**Magnetics**
image
**Strayfields**
image
**Time Integration**
image
**Domain**
image
**Utilities**
image

JonathanMaes:
It still gives the non-fatal error `ERROR: Failed to load compilation database from .\build due to: Could not auto-detect compilation database from directory ".\build"`, which I think means that MSVC does not generate `compile_commands.json`. But CUDA requires MSVC, so I don't think it is worth the hassle (i.e., installing another compiler) to fix this.

Installing the environment.yml and building mumaxplus don't give any errors, so I think it is fine as it is now. The documentation can be built on Linux anyway. And if nobody in the team would have access to a Linux machine (extraordinarily unlikely) we now know that GitHub actions can successfully build the Python wheels so if needed it should also be able to generate the documentation with the diagrams.

JonathanMaes (approved):

ilateur (changes requested):
Looks good. The one thing left to do is a bit of manual reordering of the layout, as requested by @lamoreel.

ilateur (approved):

PR #148, opened 2026-04-28
Dirichlet boundaries in anisotropic exchange

lamoreel:
This PR implements proper Dirichlet boundaries in the anisotropic exchange field calculation.

### DIFF
--- a/src/physics/atmexchange.cu
+++ b/src/physics/atmexchange.cu
@@ -102,8 +102,8 @@ __global__ void k_atmExchangeField(CuField hField,
         inter2 = interEx2.valueBetween(ridx, ridx_);
       }
     }
-    else {
-      m_ = m;
+    else { // Dirichlet boundaries
+      m_ = real3{0, 0, 0};
       a1_ = a1;
       a2_ = a2;
     }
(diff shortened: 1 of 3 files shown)


src/physics/atmexchange.cu
```diff
@@ -118,53 +118,47 @@ __global__ void k_atmExchangeField(CuField hField,
...
-    if (ridx_ != ridx) {
-      scale1 = scaleEx1.valueBetween(ridx, ridx_);
-      inter1 = interEx1.valueBetween(ridx, ridx_);
-      scale2 = scaleEx2.valueBetween(ridx, ridx_);
-      inter2 = interEx2.valueBetween(ridx, ridx_);
+    // Dirichlet boundaries
+    if (!hField.cellInGeometry(coo_) || msat.valueAt(coo_) == 0) {
+      neighbours[i] = real3{0, 0, 0};
+      Aex[i] = 0;
+      continue;
```
ilateur: I don't think this `continue` serves any purpose

ilateur (commented):
Now that I think about it, these are not truly Dirichlet boundary conditions, as the value (0,0,0) is given to the next cell, no to the boundary itself, so 1/2 cell off. 🤔

lamoreel:
> Now that I think about it, these are not truly Dirichlet boundary conditions, as the value (0,0,0) is given to the next cell, no to the boundary itself, so 1/2 cell off. 🤔

Okay, fair, but then it's still a (first order) FD approximation of a true Dirichlet boundary, since it converges to this when the cell size goes to zero. Nevertheless, the type of BC imposed here must be clear to a user. Is it?

test/test_anisotropic_exchange.py
```diff
@@ -9,20 +9,23 @@ def max_semirelative_error(result, wanted):
...
-    m_ = np.roll(m, -1, axis=3)
-    deriv[:, :, :, 0:-1] += exch[0] * (m_ - m)[:, :, :, 0:-1] / (cellsize[0] ** 2)
+    # x-direction
+    m_x_plus  = m_pad[:, :, 1:-1, 2:]
+    m_x_minus = m_pad[:, :, 1:-1, :-2]
+    # y-direction
+    m_y_plus  = m_pad[:, :, 2:, 1:-1]
+    m_y_minus = m_pad[:, :, :-2, 1:-1]
+    # center
+    m_center = m_pad[:, :, 1:-1, 1:-1]
```
JonathanMaes: Is this not just `m` again?

ilateur (approved):

JonathanMaes (approved):

PR #151, opened 2026-04-30
add the strayfield kernel to strayfield in python

DiegoDeGusem:
This PR will allow the user to access the `Field` generated by `StrayFieldKernel`. This means that we don't have to use this anymore:
```python
from _mumaxpluscpp import _demag_kernel

_demag_kernel(magnet._impl)
```
test_demag.py has also been changed such that it uses this new feature.

### DIFF
--- a/mumaxplus/strayfield.py
+++ b/mumaxplus/strayfield.py
@@ -99,3 +99,8 @@ class StrayField(FieldQuantity):
     @switch_radius.setter
     def switch_radius(self, value=-1):
         self._impl.switching_radius = value
+
+    @property
+    def kernel(self):
+        """Return the StrayFieldKernel."""
+        return self._impl.kernel
--- a/mumaxplus/strayfield.py
+++ b/mumaxplus/strayfield.py
@@ -102,5 +102,5 @@ class StrayField(FieldQuantity):
 
     @property
     def kernel(self):
-        """Return the StrayFieldKernel."""
+        """Return the stray_field_kernel."""
         return self._impl.kernel
--- a/src/physics/strayfield.hpp
+++ b/src/physics/strayfield.hpp
@@ -65,6 +65,9 @@ class StrayFieldExecutor {
   /** Return the switching radius of the executor. */
   virtual double switchingradius() const = 0;
 
+  /** Return the strayfieldkernel as a field */
+  virtual Field kernel() const = 0;
+
  protected:
   /** Source of the stray field*/
   const Magnet* magnet_;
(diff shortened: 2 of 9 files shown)


src/physics/strayfield.hpp
```diff
@@ -65,6 +65,9 @@ class StrayFieldExecutor {
   /** Return the switching radius of the executor. */
   virtual double switchingradius() const = 0;
 
+  /** Return the strayfieldkernel as a field */
+  virtual Field kernel() const = 0;
```
ilateur: I personally think it would be more logical on the C++ side to return a `StrayFieldKernel` instead of a `Field` and only specifically ask for the `.field()` inside `wrap_strayfield` for the Python side, unless I'm missing something?
(This comment applies to every `kernel()` function)

test/test_demag.py
```diff
@@ -11,7 +11,10 @@ def relative_error(result, wanted):
     return np.abs((wanted - result) / result)
 
 def demag_field_py(magnet):
-    kernel = _cpp._demag_kernel(magnet._impl, order, eps, R)
+    magnet.demag_field.order = order
+    magnet.demag_field.epsilon = eps
+    magnet.demag_field.switch_radius = R
```
ilateur: The kernel is recalculated at each of these setters. It looks clean, but for a test that needs to be as quick as possible, I would avoid this and still use the old `_cpp._demag_kernel` instead.

ilateur (changes requested):

ilateur (approved):

mumaxplus/strayfield.py
```diff
@@ -99,3 +99,8 @@ def switch_radius(self):
     @switch_radius.setter
     def switch_radius(self, value=-1):
         self._impl.switching_radius = value
+
+    @property
+    def kernel(self):
+        """Return the StrayFieldKernel."""
```
lamoreel: For some reason, we decided to use snake_case in the Python interface instead of Pascal/camel.

src/physics/strayfield.hpp
```diff
@@ -65,6 +66,9 @@ class StrayFieldExecutor {
   /** Return the switching radius of the executor. */
   virtual double switchingradius() const = 0;
 
+  /** Return the strayfieldkernel as a field */
```
lamoreel: So here PascalCase? (also line 159).

lamoreel: Also, this comment is not true.

lamoreel (approved):
Some minor comments which you may choose to ignore 😊 .

mumaxplus/strayfield.py
ilateur: Because of [PEP 8](https://peps.python.org/pep-0008/#function-and-variable-names) :yum:

PR #154, opened 2026-05-19
Use GitHub Actions to build and publish wheels for Linux and Windows

JonathanMaes:
After some trial and error, the `build-wheels.yml` GitHub Actions workflow now correctly runs when a release is manually published on GitHub, causing it to
- build Python 3.11-3.14 wheels for CUDA 12.4.1 on both Linux and Windows,
- upload a source distribution `.tar.gz` to PyPI (https://pypi.org/project/mumaxplus/),
- attach the built `.whl` files to said GitHub release (e.g., https://github.com/mumax/plus/releases/tag/v1.2.0).

The reason the wheels are not (yet?) uploaded on PyPI is because they are larger than the PyPI file upload limit (&approx;260 MB vs. 100 MB). An increased limit has been requested, we'll wait and see when (or if) that happens. Until then, mumax⁺ can be installed via:

| &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Command&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; | &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Requirements&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; | Notes |
| --- | --- | --- |
| `pip install mumaxplus` | - CUDA Toolkit installed<br>- C++ compiler installed<br>(see README for version details) | This is no different from the original method of using `pip install .` to build from source, but the repo no longer needs to be cloned and dependencies are automatically installed. |
| `pip install wheel_name.whl` | - Python 3.11-3.14 <br> - GPU with compute capability &geq;5.2 <br> - CUDA Driver &geq;550.54.15 on Linux, &geq;551.78 on Windows | Download the appropriate `wheel_name.whl` for your system from the "Assets" of a GitHub release. No other software is needed anymore. The wheels were built with CUDA 12.4.1. |

> [!WARNING]
> Since PyPI filenames can not be re-used, `mumaxplus.__version__` must be updated *before* any release.

> [!NOTE]
> The wheels are built for CUDA 12.4.1 since the Windows runners have MSVC 2022 installed, which is only supported by CUDA &geq;12.4. Since CUDA 12.4 has existed for over 2 years, I deemed it not worth the hassle to install a different MSVC on the GitHub Actions runners.


JonathanMaes:
I added a notebook to the examples folder that can be loaded and run in Google Colab. Once this PR is merged into dev, it can be found at https://colab.research.google.com/github/mumax/plus/blob/dev/examples/colab.ipynb.

Once everything is merged into master, it should be located at https://colab.research.google.com/github/mumax/plus/blob/master/examples/colab.ipynb.

lamoreel:
Thanks for this great addition! I believe this will improve the installation process for a lot of users. Before reviewing any of the changes, I want to note beforehand that this addition should be mentioned in the main repo's `readme`. The compatibility with Google Colab is also something that can be celebrated there.

.github/workflows/build-wheels.yml
```diff
@@ -0,0 +1,162 @@
+name: Build wheels
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: "Release tag (e.g., v1.2.0)"
```
ilateur, line 9: Strange description with an example?

.github/workflows/build-wheels.yml
```diff
@@ -0,0 +1,162 @@
...
+jobs:
+  build-wheels:
+    name: Build (${{ matrix.os.name }}-${{ matrix.os.version }}, CUDA ${{ matrix.cuda }}, py${{ matrix.python }})
+    runs-on: ${{ matrix.os.name }}-${{ matrix.os.version }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - {name: "ubuntu", version: "22.04"}
```
ilateur, line 21: Isn't this version a bit old now?

.github/workflows/build-wheels.yml
```diff
@@ -0,0 +1,162 @@
...
+          docker run --rm -i \
+            -v $PWD:/src \
+            -w /src \
+            nvidia/cuda:${{ matrix.cuda }}-devel-${{ matrix.os.name }}${{ matrix.os.version }} \
+            bash -euxo pipefail << 'EOF'
+              export DEBIAN_FRONTEND=noninteractive
+
+              echo "=== System info ==="
+              uname -a
+              nvcc --version || true
```
ilateur, line 56: or true?

.github/workflows/build-wheels.yml
```diff
@@ -0,0 +1,162 @@
...
+          find artifacts -type f -name "*.whl" -exec cp {} dist-gh-release/ \;
+
+          echo "=== PyPI release files ==="
+          ls -lh dist-pypi
+          
+          echo "=== GitHub release files ==="
+          ls -lh dist-gh-release
+
+      - name: Publish .tar.gz to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
```
ilateur, line 150: v1?

pyproject.toml
```diff
@@ -16,5 +68,6 @@ force-exclude = '''
 minversion = '6.0'
 addopts = '-ra -q --verbose'
 testpaths = ['test']
-[build-system]
-requires = ["setuptools", "cmake"]
+
+[tool.setuptools.dynamic]
+version = { attr = "mumaxplus.__version__" }
```
ilateur, line 73: nice

examples/colab.ipynb
```diff
@@ -0,0 +1,212 @@
...
+        "    import requests\n",
+        "except ImportError:\n",
+        "    pass\n",
+        "else:\n",
+        "    # Determine which wheel to use\n",
+        "    from platform import python_version_tuple\n",
+        "    major, minor, patch = python_version_tuple()\n",
+        "    PYVERSION = major + minor\n",
+        "    # Download the mumax+ wheel\n",
+        "    URL = f\"https://github.com/mumax/plus/releases/download/v{MUMAXPLUS_VERSION}/mumaxplus-{MUMAXPLUS_VERSION}-cp{PYVERSION}-cp{PYVERSION}-manylinux_2_34_x86_64.whl\"\n",
```
ilateur, line 85: What's the point of printing `PYVERSION` twice in the file name `-cp{PYVERSION}-cp{PYVERSION}`?

ilateur (commented):
I haven't been able to test these new installation methods (yet) and the build-wheels workflow mostly goes over my head, but it looks pretty good!

Perhaps you could add a little project description to the [mumaxplus Pypi package](https://pypi.org/project/mumaxplus/#description), which also refers to the [GitHub page](https://github.com/mumax/plus) and the [website](https://mumax.github.io/plus/)?

.github/workflows/build-wheels.yml
JonathanMaes: The wheels are forwards compatible, so generating them on 24.04 gives errors when installing them on 22.04.

.github/workflows/build-wheels.yml
JonathanMaes: `|| true` makes sure this line can not give an error and thereby interrupt the workflow before the actual compilation starts.

.github/workflows/build-wheels.yml
JonathanMaes: Only the major version number is specified in `.yml`, and this package is currently at 1.14.0. If any breaking changes are introduced in this action in the future, they should create a version 2.0.0, which we hereby avoid.

examples/colab.ipynb
JonathanMaes: The filename of a wheel turns out to be extremely important and follows a fixed format, otherwise it will not install: `{distribution}-{version}(-{build tag})?-{python tag}-{abi tag}-{platform tag}.whl`.

The two you are talking about are `python tag` and `abi tag`. The first specifies which version of the Python API (i.e. the syntax) the wheel should work on. Since most packages are pure Python, they only need to generate one wheel, and backwards compatibility of the Python syntax will take care of the rest. However, since `mumaxplus` heavily relies on CUDA and C, we must ensure that also the ABI (application binary interface) matches. Since the ABI is not guaranteed to be stable across Python versions, we must specify it explicitly (in most packages' wheels, this tag would be `none`). Hence, we end up with `-cp{PYVERSION}` twice. It would probably also work with `-cp{PYVERSION}-none`, but better safe than sorry.

.github/workflows/build-wheels.yml
JonathanMaes: There are two ways to start this workflow: either by publishing a release (`release: types: [published]`), or by manually starting it (`workflow_dispatch`). In the latter case, a GitHub release will be created automatically, since otherwise the built wheels will only be temporary files hidden inside the GitHub action. For this, the tag to which this release should be attached is needed. The `description:` here describes that input value in the GitHub GUI, I believe.

However, I don't think the code being built actually uses this tag, and it also does not affect the version number uploaded to PyPI since that is governed by the `__version__`. So I think manually running the workflow would just checkout the latest master commit, and publish a new release linked to whatever tag you specify in this input.

JonathanMaes:
> Perhaps you could add a little project description to the [mumaxplus Pypi package](https://pypi.org/project/mumaxplus/#description), which also refers to the [GitHub page](https://github.com/mumax/plus) and the [website](https://mumax.github.io/plus/)?

This will be added automatically when the workflow is run. Currently, the mumaxplus PyPI page refers to version 1.2.0, which did not yet use `pyproject.toml` etc.

However, the "test" version I initially published (https://pypi.org/project/mumaxplus/1.2.0.dev1/) does have that info, so upon the release of 1.2.1 that is what the webpage will look like.

ilateur:
I successfully installed mumax+ via every method suggested in the README, except for `Installing pre-built wheels`, because monster5 is ancient. A few thoughts/suggestions:
- Perhaps it would be helpful to have some guidance on which `.whl` you should download and install. win/linux is obvious, but that `cp313` means python 3.13 is slightly less obvious to me.
- Additionally, knowing some restrictions up front would also be nice, like if I know monster5 will fail, I should not try that method. It has to do with the `34` in the name, but I do not know enough to formulate the exact specification. Perhaps a little drop-down tip about `pip debug` would help?
- Both `Installing pre-built wheels` and `Installing from source: Installing a stable release from PyPI` assume `pip` (and Python for pre-built) is already installed. Perhaps it would be nice to have a little help setting up the (conda) environment for noobs? For example:
```bash
conda create --name mumaxplus
conda activate mumaxplus
conda install python
```

DiegoDeGusem (commented):
I have successfully installed mumax⁺ with every method including the wheels for each Python version.

JonathanMaes:
> * Both `Installing pre-built wheels` and `Installing from source: Installing a stable release from PyPI` assume `pip` (and Python for pre-built) is already installed. Perhaps it would be nice to have a little help setting up the (conda) environment for noobs? For example:
> ```shell
> conda create --name mumaxplus
> conda activate mumaxplus
> conda install python
> ```

Conda is not necessary, except in the case of `Installing from source: Installing a custom mumax⁺` where one should use `environment.yml`, but there the readme already links to the conda downloads page. For the other cases, we assume Python 3.11-3.14 is already installed (which usually includes pip). I don't think it is our responsibility to teach people how to use Python environments, the readme is already getting rather cluttered.

Your other comment I tried to address by adding a tip about `pip debug -v`.

ilateur (approved):

DiegoDeGusem (approved):

lamoreel (changes requested):
I find the "Installation" section of the main `README` too cluttered as it is. I suggest adding "Option x:" to each of the three ways of installing mumax+ to make this more structured. Perhaps drop-down?
Since this file is getting longer and longer, is it possible to place a hyperlinked TOC at the beginning to make this even more user friendly?

lamoreel (approved):

lamoreel (approved):

Issue #27, opened 2024-10-04
Noise explodes


lamoreel:
Noise explodes, ultimately resulting in `nan`s, but **only when `N` is odd**.
Minimal working example:
```python
from mumaxplus import Ferromagnet, Grid, World

size = 5e-10
N = 1
nx, ny, nz = N, N, 1
grid = Grid((nx, ny, nz))
world = World(cellsize=(size, size, size))

magnet = Ferromagnet(world, grid)
magnet.msat = 400e3
magnet.alpha = 0.005

magnet.temperature = 1

world.timesolver.steps(5)

print(magnet.thermal_noise())
```

PR #155, opened 2026-05-22
Noise QOL

lamoreel:
This PR adds some (minor) QOL features in the thermal noise calculation:

- The seed of the RNG is now settable through a new `Ferromagnet` property `thermal_seed`.
- The CUDA RNG is only compatible with an even number of grid cells. As suggested by @jplauzie in #27, the random field in the case of an odd number of grid cells `N` is now calculated for `N + 1` cells. The last value is discarded in the actual noise calculation.
- The RNG can now be reset using `magnet.reset_noise_generator()`.


test/test_thermal_seed.py
```diff
@@ -0,0 +1,29 @@
+import numpy as np
+from mumaxplus import Ferromagnet, Grid, World
+
+class TestThermalSeed:
+    @staticmethod
+    def generate_noise(seed=None):
+        world = World((1e-9, 1e-9, 1e-9))
+        magnet = Ferromagnet(world, Grid((16, 16, 4)))
```
JonathanMaes: An odd number of cells (e.g., `(15, 15, 3)`) could be used here just as an extra check.

src/physics/thermalnoise.cu
```diff
@@ -57,10 +57,14 @@ Field evalThermalNoise(const Ferromagnet* magnet) {
...
+
+  real* tmp;
+  cudaMalloc(&tmp, (N + (N % 2)) * sizeof(real));
   for (int c = 0; c < 3; c++) {
-    generateRandNormal(magnet->randomGenerator, noise.device_ptr(c), N, mean,
-                       stddev);
+    generateRandNormal(magnet->randomGenerator, tmp, N + (N % 2), mean, stddev);
+    cudaMemcpy(noise.device_ptr(c), tmp, N * sizeof(real), cudaMemcpyDeviceToDevice);
   }
+  cudaFree(tmp);
```
ilateur: I am worried that `cudaMalloc`, `cudaMemcpy` and `cudaFree` are (much) slower than the old `noise.device_ptr` (essentially twice as many `Field` creations). 😕 Could this 'before and after' be tested? Simulations with temperature are already slow enough as it is.

If it is noticeably slower, perhaps `tmp` could only be created in case of an odd number of cells, and otherwise use noise?

ilateur (commented):
I thought of an extra test (I think). Make a magnet with e.g. (N, 1, 1) cells and (N+1, 1, 1) cells with N an odd number. Then keep the seed consistent and ask for thermal noise. Those first N numbers should be exactly the same, right?

src/physics/thermalnoise.cu
lamoreel: I timed standard problem 4 at 500K, and the difference is negligible. However, you're very right in questioning this approach and this may cause a significant memory overhead for large simulations. An additional check will be added before `tmp` creation.

DiegoDeGusem (approved):

src/physics/thermalnoise.cu
```diff
@@ -58,13 +58,20 @@ Field evalThermalNoise(const Ferromagnet* magnet) {
...
-  cudaMalloc(&tmp, (N + (N % 2)) * sizeof(real));
-  for (int c = 0; c < 3; c++) {
-    generateRandNormal(magnet->randomGenerator, tmp, N + (N % 2), mean, stddev);
-    cudaMemcpy(noise.device_ptr(c), tmp, N * sizeof(real), cudaMemcpyDeviceToDevice);
+  // CUDA RNG is only compatible with even N.
+  if (N % 2) {
+    real* tmp;
+    cudaMalloc(&tmp, (N + (N % 2)) * sizeof(real));
+    for (int c = 0; c < 3; c++) {
+      generateRandNormal(magnet->randomGenerator, tmp, N + (N % 2), mean, stddev);
```
ilateur: (This could now safely become N+1 😋)

ilateur (approved):

JonathanMaes (approved):

PR #158, opened 2026-06-03
Sensible noise

ilateur:
# The Idea
The idea is to expand the sensible time step calculation with thermal noise
$dt = f / \max{(|\tau|)}$
with $f$ a sensible factor. This slightly complicates things, because the thermal noise torque depends on the time step. Take $n$ as all thermal noise and prefactors except the time step $dt$, then
$\tau \rightarrow \tau + \frac{1}{\sqrt{dt}} n$
To deal with the maximum and norm, I used
$\max{\left( |\tau + \frac{1}{\sqrt{dt}} n | \right)} \quad \le \quad \max{(|\tau|)} + \frac{1}{\sqrt{dt}} \max{(|n|)}$
This becomes a second order equation to solve.

One bug I encountered was when noise dominates over torque, which can easily happen with minimized magnetization, that the sensible time step would become identically 0 due to numerical limitations. In this limit I used a different function.

# Graphing Behavior
To check this visually, we can move to dimensionless parameters. We can define dimensionless temperature
$x = N^2 / (f \mathcal{T})$
with $N$ the max norm of noise (sqrt of temperature) and $\mathcal{T}$ the max norm of torque in the system. We can also define a dimensionless time step.
$y = dt * \mathcal{T} / f$
Keeping $\mathcal{T}$ and $f$ constant (in a skyrmion system) and varying temperature, we get the following time step estimates.
image
At $x=0$ we get the old sensible time step formula ($y=1$). Adding any noise decreases this time step, where eventually noise will dominate.
We can see that this behaves sensibly. Inspired by all cases of this figure, I added the tests.

# Consistency
One thing that bothers me is that the sensible time step is now random, because taking the max norm of the noise is random. It is possible to estimate the expected maximum norm of the thermal noise. First, you can simply take the maximum pre-factor of all physical variables. For the noise itself, each norm of 3 random Gaussians is distributed as a $\chi_3$ (or Maxwell-Boltzmann) distribution. The maximum of a collection of these numbers depends on the number of cells in the geometry. Using Extreme Value Theory, the expected value of this maximum can be calculated in the limit of a large number of cells. This works fairly well, as seen in the following figure.
image

## The Problem
The problem in the code base, is that this would require each `DynamicEquation` to keep track of an extra `QuantityEvaluator` , simply for the `Timesolver` to estimate the noise about once per simulation if temperature is non-zero. This felt so stupid, that I did not implement it.


src/core/timesolver.cpp
```diff
@@ -37,13 +37,38 @@ RKmethod TimeSolver::getRungeKuttaMethod() {
...
+    }
+  }
+
+  if (maxNoise == 0) {
+    // Sensible timestep cannot be calculated if torque is zero
+    if (maxRhs == 0) return sensibleTimestepDefault();
+    // with RHS but no noise
+    return sensibleFactor() / maxRhs;
+  }
+  // negligable RHS compared to noise
```
JonathanMaes: negligible*

JonathanMaes (approved):

src/core/timesolver.cpp
```diff
@@ -37,13 +37,38 @@ RKmethod TimeSolver::getRungeKuttaMethod() {
...
+  real smallNumber = 1e-3;
```
lamoreel, line 66: Does this value have any meaning? Is this "small enough"? Is it extracted from your skyrmion example? Is this limit always applicable?

lamoreel (approved):

src/core/timesolver.cpp
ilateur: It does not have any inherent meaning. "Small enough" can be justified via a Taylor expansion:
$y = (\sqrt{1 + \frac{4}{x}} - 1)^2 \frac{x}{4} \approx (1 + \frac{2}{x} - 1)^2 \frac{x}{4} = \frac{1}{x}$
Here the "measure of appropriateness" is how small $\frac{2}{x} = 2 \frac{f \mathcal{T}}{N^2}$ is. I felt like 1e-3 was a good middle ground between large (1) and numerical noise (1e-7). This is (in my opinion) confirmed in the first graph above, where the left and right functions join smoothly over the dashed line. That graph has been made via my skyrmion example, but it should be situation agnostic due to the "dimensionlessness".

Perhaps I will expand my code comment a bit.

PR #159, opened 2026-06-05
New shape to load 3D objects

JonathanMaes:
This PR adds a new `Shape`, which can load a 3D object from STL, PLY, OBJ or GLTF/GLB files.
This **`ObjShape`** takes quite a few arguments:
- The name of the file to be loaded
- Exactly 2 of the following parameters to define the object's bounding box: `min_point`, `max_point`, `center`, `scale` and `size`. The last two of these parameters are mutually exclusive.
- Optional:
  - `keep_aspect` (default `False`): expand to the bounding box defined by the previous parameters without distortion.
  - `repair` (default `False`): try to make the shape watertight if it is not. A simply-connected object is assumed. 
  - `rotate_z_up` (default `True`): most 3D objects use the y-axis as the vertical direction, while in mumax this is the z-axis.

### Example: magnetization of a teapot

```python
from mumaxplus import Ferromagnet, World, Grid
from mumaxplus.util import show_field_3D, ObjShape

N, L = 64, 512e-9
shape = ObjShape("teapot.obj", (0, 0, 0), size=(L, L, L), keep_aspect=True, rotate_z_up=True)
world = World(cellsize=(L/N, L/N, L/N))
magnet = Ferromagnet(world, Grid((N, N, N)), geometry=shape)
magnet.msat = 800e3
magnet.aex = 13e-12
magnet.relax()
show_field_3D(magnet.magnetization, enable_quiver=False)
```
image

### Other changes
- Renamed the 'show' optional dependency group to '3D' and added `trimesh` and `networkx` to it. These optional dependencies should therefore now be installable alongside mumax⁺ with `pip install mumaxplus[3D]`
- Initializing a geometry now uses the `shape_func` directly if it is already vectorized, otherwise it falls back to `_np.vectorize` like before
- Added this shape to `examples/shape.ipynb`, where the quintessential teapot is loaded
- Fixed formatting in `Torus` and `repeat` documentation


mumaxplus/magnet.py
```diff
@@ -64,7 +64,10 @@ def _get_mask_array(input, grid, world, input_name):
             x, y, z = [(grid.origin[i] + idxs[i]) * cs[i] for i in [0, 1, 2]]
 
             # evaluate the input function for each position in this meshgrid
-            return _np.vectorize(input, otypes=[T])(x, y, z)
+            try:
+                return input(x, y, z) # If input function is already vectorized
+            except Exception:
+                return _np.vectorize(input, otypes=[T])(x, y, z)
```
ilateur, line 70: I had hoped for a fancy solution... but I guess `try except` is the Pythonic way 😔

examples/shape.ipynb
ilateur, line 1: Could you add this to `docs/tutorial/shapes.rst` too, including an image?

mumaxplus/util/shape.py
```diff
@@ -752,6 +756,98 @@ def shape_func(x, y, z):
...
+        fname : string
+            Filename of the 3D object to use.
+        min_point : tuple[float] of size 3
+            Smallest x, y and z coordinates of the object's bounding box.
+        max_point : tuple[float] of size 3
+            Largest x, y and z coordinates of the object's bounding box.
+        center : tuple[float] of size 3
+            Center of the object's bounding box.
+        scale : float or tuple[float] of size 3
+            Scaling factor to apply. If a tuple, scaling for each axis.
```
ilateur, line 784: I am a bit confused about the use of `scale`. The other parameters seem much more useful to me. Also all `Shape`s already have a `scale` method.

mumaxplus/util/shape.py
```diff
@@ -752,6 +756,98 @@ def shape_func(x, y, z):
...
+            mumax, where the Z-axis represents the vertical direction.
+        """
+        import pyvista as pv # Only attempt import if this shape is used, since this is an optional dependency
+        import trimesh
+        
+        ## Load the mesh in the correct orientation
+        mesh = trimesh.load(fname)
+        if repair: trimesh.repair.fill_holes(mesh)
+        if rotate_z_up: # Rotate Y-up to Z-up (swap Y- and Z-axes)
+            rotation_matrix = trimesh.transformations.rotation_matrix(_np.pi/2, [1, 0, 0]) # 90° rotation around X-axis
```
ilateur: (swapping axes and rotating is different)

mumaxplus/util/shape.py
```diff
@@ -752,6 +756,98 @@ def shape_func(x, y, z):
...
+            mesh.apply_transform(rotation_matrix)
+        mesh_size = mesh.bounds[1] - mesh.bounds[0]
+        
+        ## Parse the positioning arguments
+        toarray = lambda arg: _np.asarray(arg) if arg is not None else None
+        min_point, max_point, center, scale, size = toarray(min_point), toarray(max_point), toarray(center), toarray(scale), toarray(size)
+        if size is not None:
+            if scale is not None: raise ValueError("You can either specify 'scale' or 'size', not both at the same time.")
+            scale = size / mesh_size
+        if (min_point is None) + (max_point is None) + (center is None) + (scale is None) != 2:
```
ilateur: You are counting how many arguments have *not* been set, which feels a bit backwards to me, but because 4=2+2 it works out.

mumaxplus/util/shape.py
```diff
@@ -752,6 +756,98 @@ def shape_func(x, y, z):
...
+            raise ValueError("Exactly 2 arguments of 'min_point', 'max_point', 'center' and 'scale'/'size' should be provided.")
+        match (min_point, max_point, center, scale): # Reduce parameters to just (min_point, max_point)
+            case (_, _, None, None): pass
```
ilateur, line 821: I did not know that match case could be used with underscores in this way. Good to know!

ilateur (changes requested):

mumaxplus/util/shape.py
JonathanMaes: If your .obj would for some reason already be nanometre-sized, you could pass `scale=1` rather than having to calculate the bounding box corners/size yourself.

mumaxplus/magnet.py
JonathanMaes: This will throw an error very quickly if the input function is not yet vectorized, likely a negligible time loss compared to calculating whether each cell is inside the geometry.

mumaxplus/util/shape.py
ilateur: Hmm I suppose so :thinking: this could also work by only specifying the min_point or center and nothing more, but then the clear "rule of 2" disappears

JonathanMaes: Indeed, but then what if you would want to know what the magnetization of a half-sized version of your perfectly pre-sized .obj is? Rather than `scale=0.5`, one would then have to calculate the bounding box 😔
(ignore the near-zero probability of such a situation occurring for anyone, ever)

mumaxplus/util/shape.py
```diff
@@ -752,6 +756,98 @@ def shape_func(x, y, z):
...
+            to fill said bounding box completely along all axes. Hence, when
+            keep_aspect is True, the values passed to min_point, max_point and
+            scale/size may not fully correspond to the resultant bounding box.
+        repair : bool (default=False)
+            Attempts to interpret the object as a single solid, attempting
+            basic repairs in case the mesh is not watertight.
+        rotate_z_up : bool (default=True)
+            If your .obj file uses the Y-axis as the vertical direction, you
+            can set `rotate_z_up=True` to rotate your 3D object correctly in
+            mumax, where the Z-axis represents the vertical direction.
```
lamoreel: I prefer "out-of-plane direction" over "vertical direction" in this context.

mumaxplus/util/shape.py
```diff
@@ -752,6 +756,98 @@ def shape_func(x, y, z):
...
+        
+        ## Parse the positioning arguments
+        toarray = lambda arg: _np.asarray(arg) if arg is not None else None
+        min_point, max_point, center, scale, size = toarray(min_point), toarray(max_point), toarray(center), toarray(scale), toarray(size)
+        if size is not None:
+            if scale is not None: raise ValueError("You can either specify 'scale' or 'size', not both at the same time.")
+            scale = size / mesh_size
+        if (min_point is not None) + (max_point is not None) + (center is not None) + (scale is not None) != 2:
+            raise ValueError("Exactly 2 arguments of 'min_point', 'max_point', 'center' and 'scale'/'size' should be provided.")
+        match (min_point, max_point, center, scale): # Reduce parameters to just (min_point, max_point)
```
lamoreel: Is this too strict? What if all 4 arguments are provided, but also consistent with one another? Is a warning sufficient (e.g. when ignoring certain input values)?

lamoreel (approved):

mumaxplus/util/shape.py
ilateur: There is not really a "plane" to speak of with 3D shapes

lamoreel: Okay :shrug:.

mumaxplus/util/shape.py
JonathanMaes: True, but if the user knows all 4 arguments, they might just as well choose 2 of those. 

Checking consistency would make the argument parsing even longer than it already is. For instance, the following would have to be checked, which looks rather confusing:
```python
numNotNone = (min_point is not None) + (max_point is not None) + (center is not None) + (scale is not None)
if numNotNone < 2:
    raise ValueError("At least 2 arguments of 'min_point', 'max_point', 'center' and 'scale'/'size' should be provided.")
elif numNotNone == 3:
    if min_point is None:
        assert _np.allclose(max_point - center, scale*mesh_size/2)
        max_point = None
    elif max_point is None:
        assert _np.allclose(center - min_point, scale*mesh_size/2)
        min_point = None
    elif center is None: assert _np.allclose(max_point - min_point, scale*mesh_size)
    elif scale is None: assert _np.allclose(max_point - center, center - min_point)
elif numNotNone == 4:
    assert _np.allclose(max_point - center, center - min_point) and _np.allclose(max_point - min_point, scale*mesh_size)
    scale = center = None
```

lamoreel: Okay, two it is :)

DiegoDeGusem (approved):
Nice!
Image

ilateur (approved):

PR #161, opened 2026-06-17
Moving simulation window for domain wall motion

lamoreel:
This pull request adds the possibility for a moving simulation window, albeit not very general. The functionality is implemented for
- Magnets with no geometry
- Magnets with no regions
- Only domain wall motion is supported (1D)

The code base is kept general as to easily extend it to other cases (e.g. 2D skyrmion motion) in later PRs. The (ancient) implementation on the `moving-window` branch already contains some groundwork for a more general implementation, in the case we or the community would need this in the future.


mumaxplus/world.py
```diff
@@ -334,3 +348,51 @@ def unset_pbc(self):
...
+            origin = np.array(self.bounding_grid.origin) * np.array(self.cellsize)
+        return Window(self._impl.window, origin)
+
+    def center_domain_wall(self, comp, axis=None):
+        """Move the simulation window along with the domain wall. This function
+        should be called before the `TimeSolver`.
+
+        Note
+        ----
+        The domain wall should be centered already for this function to work properly.
```
ilateur, line 363: Is there no way to do this automatically when called?

mumaxplus/world.py
```diff
@@ -334,3 +348,51 @@ def unset_pbc(self):
...
+            grid cells is chosen.
+        """
+        if comp not in (0, 1, 2):
+            raise ValueError("The component `comp` should be 0 (x), 1 (y) or 2 (z).")
+
+        if len(self.magnets) != 1:
+            raise RuntimeError("The moving window functionality only works if exactly 1 magnet exists.")
+
+        magnet = list(self.magnets.values())[0]
+        if not np.any(magnet.geometry) or np.any(magnet.regions):
```
ilateur: I think this if-statement is incorrect. "not any geometry" tests whether or not the geometry is completely empty (which should never be the case). I believe `not np.all(magnet.geometry)` would be the correct check.

mumaxplus/world.py
```diff
@@ -334,3 +348,51 @@ def unset_pbc(self):
...
         """
         self._impl.unset_pbc()
+
+    @property
+    def window(self) -> Window:
+        """Simulation window for this world."""
+        origin = None
+        if self.bounding_grid.origin != (0, 0, 0):
+            origin = np.array(self.bounding_grid.origin) * np.array(self.cellsize)
+        return Window(self._impl.window, origin)
```
ilateur: Why does `origin` need to be determined here and is it not included in `self._impl.window`?

mumaxplus/window.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def _check_boundary(self, boundary):
+        if not isinstance(boundary, int) or boundary not in (0, 1):
+            raise ValueError(f"Invalid boundary: {boundary}. Must be one of (0, 1).")
+
+
+    def insert_magnetization(self, boundary, value):
+        """
+        Set magnetization value at a given boundary.
+        If none are given (or set to zero), the current edge value is used.
```
ilateur: "If none are given" does not seem to be supported? (With this wording, I would expect that literally passing `None` would work.)

mumaxplus/window.py
```diff
@@ -0,0 +1,48 @@
...
+            Boundary index: 0 = Left/Bottom, 1 = Right/Top
+        value : tuple of 3 floats
+            Magnetization vector (x, y, z)
+        """
+        self._check_boundary(boundary)
+        self._impl.insert_magnetization(boundary, value)
+
+    @property
+    def position(self):
+        """Returns the current position of the simulation window."""
```
ilateur: How is the position of a window determined? The coordinate of the lower left corner?

mumaxplus/window.py
```diff
@@ -0,0 +1,48 @@
...
+        return self._impl.position
+
+    @property
+    def velocity(self):
+        """Returns the current velocity of the simulation window."""
```
ilateur: In m/s?

mumaxplus/window.py
```diff
@@ -0,0 +1,48 @@
...
+        return self._impl.velocity
+
+    @property
+    def total_shift(self):
+        """Returns the total amount shifted by the simulation window."""
```
ilateur: number of cells or meters?

src/bindings/wrap_window.cpp
```diff
@@ -0,0 +1,22 @@
...
+    py::class_<Window>(m, "Window")
+        .def(py::init<>())
+        .def("_set_origin", &Window::setOrigin, py::arg("origin"))
+        .def("insert_magnetization", [](Window& self, int side, real3 value) {
+            self.setMagValue(static_cast<Boundary>(side), value);
+        }, py::arg("side"), py::arg("value"))
+
+        .def("position", &Window::position)
+        .def("velocity", &Window::velocity)
+        .def("total_shift", &Window::GetTotalShift);
```
ilateur: Here they are defined as functions, but in `window.py` they are defined as properties (which is better imo). I think here it should be changed to `.def_property_readonly`, otherwise you still need to add parentheses like `world.window.position()`, which shouldn't be the case.

src/bindings/wrap_window.cpp
```diff
@@ -0,0 +1,22 @@
+#include <pybind11/functional.h>
+#include <pybind11/numpy.h>
+
+#include <memory>
```
ilateur: I don't think functional, numpy and memory are needed here.

mumaxplus/world.py
```diff
@@ -334,3 +348,51 @@ def unset_pbc(self):
...
+
+    def center_domain_wall(self, comp, axis=None):
```
ilateur, line 357: Seeing as there are a lot of restrictions to using this function, should we add `ext_` before the name like it is done in mumax3? Or don't we do that here?

mumaxplus/world.py
```diff
@@ -334,3 +348,51 @@ def unset_pbc(self):
...
+        if comp not in (0, 1, 2):
+            raise ValueError("The component `comp` should be 0 (x), 1 (y) or 2 (z).")
+
+        if len(self.magnets) != 1:
+            raise RuntimeError("The moving window functionality only works if exactly 1 magnet exists.")
+
+        magnet = list(self.magnets.values())[0]
+        if not np.any(magnet.geometry) or np.any(magnet.regions):
+            raise RuntimeError("The moving window functionality doesn't work well with geometry"
+                               " or regions as of yet.")
```
ilateur, line 394: What about nonuniform parameters defined without regions? Fairly cumbersome to test this... maybe a `Warnings` section in the docstring?

src/physics/mumaxworld.cpp
```diff
@@ -327,3 +329,42 @@ void MumaxWorld::unsetPBC() {
...
+  if (magnets_.size() > 1)
+    throw std::runtime_error("Moving the simulation window is only possible when only one "
+                             "magnet lives in the world.");
+  if (magnets_.size() < 1)
+    throw std::runtime_error("Moving the simulation window is not possible when there is no "
+                             "magnet in the world.");
+
+  Magnet* magnet = magnets_.begin()->second;
+  timesolver_->setPostStepFunction([this, magnet, comp, axis]() {
+  const Field& mag = magnet->asHost() ? magnet->asHost()->sublattices()[0]->magnetization()->field()
```
ilateur, line 346: I wonder if sub1 magnetization or neel vector (or the NCAFM equivalent) is the better choice? In most cases this is basically the same... but what about those other cases 🤔

src/core/window.hpp
```diff
@@ -0,0 +1,45 @@
...
+   real3 GetTotalShift() const { return total_dist_; }
+
+  private:
+   // Values to insert at the boundaries
+   std::array<real3, 2> magValues_;
+
+
+   // Current excitation position and velocity
+   real3 ext_pos_;
+   real3 ext_vel_;
```
ilateur: Does `ext_` stand for "exTcitation", or extension? If extension, I feel like the whole `Window` class already is the extension so that is a bit redundant.

src/core/window.hpp
```diff
@@ -0,0 +1,45 @@
...
+  Right
+};
+
+class Window {
+  public:
+   Window();
+   ~Window() = default;
+
+   // Set origin of the simulation window
+   void setOrigin(real3 origin) { ext_pos_ = origin; }
```
ilateur: I am quite confused with the meaning of `ext_pos_`. When set, it seems to be the origin of the simulation window, similar to the grid origin, but it meters. But the description says excitation position, which would be the center of the magnet?

src/core/window.hpp
```diff
@@ -0,0 +1,45 @@
...
+   void setMagValue(Boundary side, real3 value) { magValues_[idx(side)] = value; }
+   // Get values to be inserted at the boundaries
+   std::array<real3, 2> getMagValues() { return magValues_; }
+
+   Field centerOnExcitation(const Field& field, int dir, int axis, int comp);
+
+   // Get total amount shifted
+   real3 position() const { return ext_pos_; }
+   real3 velocity() const { return ext_vel_; }
+   real3 GetTotalShift() const { return total_dist_; }
```
ilateur: (`get` with noncapital letter)

src/core/window.hpp
```diff
@@ -0,0 +1,45 @@
...
+   // Values to insert at the boundaries
+   std::array<real3, 2> magValues_;
+
+
+   // Current excitation position and velocity
+   real3 ext_pos_;
+   real3 ext_vel_;
+
+   // Total distance travelled by excitation
+   real3 total_dist_;
```
ilateur: What is the point of keeping these as `real3` if you can only ever shift in a single direction (using `real` seems simpler)? Or is preemptively more general for potentially shifting in 2 directions like with skyrmions? 🤔

src/core/window.cpp
```diff
@@ -0,0 +1,21 @@
...
+    }
+
+Field Window::centerOnExcitation(const Field& field, int dir, int axis, int comp) {
+    real3 cellsize = field.world()->cellsize();
+    real cs = (&cellsize.x)[axis];
+
+    // Multiply dir by -1 because the field is moved to dir iff the window is moved to -dir
+    (&ext_pos_.x)[axis] += -1. * dir * cs;
+    (&ext_vel_.x)[axis] = -1. * dir * cs / field.world()->timesolver().timestep();
+    (&total_dist_.x)[axis] += cs;
```
ilateur: wait can dir only be 1 and -1 or can it be more? If more, this needs to be abs(dir) *cs

src/physics/mumaxworld.cpp
```diff
@@ -327,3 +329,42 @@ void MumaxWorld::unsetPBC() {
...
+    if (dir != 0) {
+      // Shift magnetization
+      auto shifted = window_->centerOnExcitation(mag, dir, axis, comp);
+
+      // Multi-sublattice systems
+      if (auto host = magnet->asHost()) {
+        auto sub0 = host->sublattices()[0];
+        sub0->magnetization()->set(shifted);
+        for (auto sub : host->getOtherSublattices(sub0)) {
+          auto shifted = window_->centerOnExcitation(sub->magnetization()->field(), dir, axis, comp);
```
ilateur, line 362: Each time `centerOnExcitation` is called, the window position is shifted internally, meaning for AFM and NCAFM it will 'move' twice/thrice as much.

ilateur, line 362: `centerOnExcitation` calls `shift` internally, to which `magValues` are passed. If these are set by the user in accordance with sub1, then the edges will be wrong for the other sublattices (I think?).

src/core/window.hpp
ilateur, line 1: I am a bit confused about what the role of the `Window` class is exactly. It mostly seems to be something akin to `Grid`: very passive, only holding onto a bit of data like position, velocity, magValues, etc., while the active shifting functionality seems to be split off from it into individual functions in `shift.hpp` (calculation) and within `postStep` defined in `MumaxWorld` (setting). 
That seems fine to me, but then `Window` also has `centerOnExcitation` which seems slightly too "powerful" for this class, while also in itself not achieving much (mostly a wrapper of `shift` and returning back the result).
I feel like a slight restructuring of "rights" is needed, if I am somewhat understandable?

src/core/shift.hpp
```diff
@@ -0,0 +1,8 @@
+#pragma once
+
+#include "datatypes.hpp"
+
+class Field;
+
+int calculateShiftDirection(const Field& field, real3 leftValue, real3 rightValue, int comp, int axis);
+Field shift(const Field& field, int dir, int axis, int comp, real3 leftValue, real3 RightValue);
```
ilateur: (Same arguments (except dir), but completely different order 🤔)

src/core/shift.cu
```diff
@@ -0,0 +1,86 @@
+#include "cudaerror.hpp"
+#include "cudalaunch.hpp"
+#include "cudastream.hpp"
+#include "field.hpp"
+#include "shift.hpp"
+
+__global__ void k_getSignOfOuterValue(int* result, CuField f, int comp, int axis) {
+    const int idx = threadIdx.x + blockIdx.x * blockDim.x;
+
+    if (idx == 0) {
```
ilateur: launching a global reduction kernel to obtain a single value from a single position feels very much like overkill. Maybe a direct `cudaMemcpy` to an int (reference/pointer) is possible?

src/core/shift.cu
```diff
@@ -0,0 +1,86 @@
...
+        GpuBuffer<int> d_result(1);
+        cudaLaunchReductionKernel(k_getSignOfOuterValue, d_result.get(), field.cu(), comp, axis);
+
+        int result;
+        checkCudaError(cudaMemcpyAsync(&result, d_result.get(), 1 * sizeof(int),
+                                       cudaMemcpyDeviceToHost, getCudaStream()));
+        (&leftValue.x)[comp] = real(result);
+    }
+    if (abs(av) > tolerance)
+        return - sgn((&leftValue.x)[comp]) * sgn(av);
```
ilateur: This deduction is unnecessary if it turns out that `abs(av) <= tolerance`, so I would insert the first if-statement into the second?

src/core/shift.cu
```diff
@@ -0,0 +1,86 @@
...
+    real3 value;
+
+    const int3 src = grid.index2coord(idx);
+
+    int3 direction{0, 0, 0};
+    (&direction.x)[axis] = dir;
+
+    int3 dst = src - direction;
+
+    if ((&dst.x)[axis] >= 0 && (&dst.x)[axis] < (&gridsize.x)[axis] && field.cellInGeometry(grid.coord2index(dst)))
```
ilateur: Isn't this long statement the same as `if (field.cellInGeometry(dst))`? Also `grid.coord2index(int3 coo)` is unsafe if the `coo` is outside the grid.

src/core/shift.cu
```diff
@@ -0,0 +1,86 @@
...
+        value = field.vectorAt(dst);
```
ilateur: I feel like `src` and `dst` are inversely named? You pick the value from `dst` and place it at `src` (`idx`).

test/test_moving_window.py
```diff
@@ -0,0 +1,230 @@
...
+    def slc(start, stop):
+        s = [slice(None), slice(None), slice(None)]
+        s[2 - axis] = slice(start, stop)
+        return tuple(s)
+
+    m = np.zeros((3, nz, ny, nx))
+
+    m[comp][slc(None, center - width)] = 1  # 'left' domain
+    m[comp][slc(center + width, None)] = -1 # 'right' domain
+    m[(comp + 1) % 3][slc(center - width, center + width)] = 1 # domain wall
```
ilateur, line 44: Why not use the `twodomain` config? (Or perhaps the Walker ansatz if you're feeling fancy)

test/test_moving_window.py
```diff
@@ -0,0 +1,230 @@
...
+    def test_non_integer_boundary(self):
+        world, _ = make_ferromagnet()
+        with pytest.raises((ValueError, TypeError, Exception)):
+            world.window.insert_magnetization(1.5, (1, 0, 0))
+
+    def test_invalid_boundary(self):
+        world, _ = make_ferromagnet()
+        for b in (-1, 4, 1.5):
+            with pytest.raises((ValueError)):
+                world.window.insert_magnetization(b, (1, 0, 0))
```
ilateur, line 60: this is the same as the previous 3 tests?

test/test_moving_window.py
```diff
@@ -0,0 +1,230 @@
...
+
+    def test_origin(self):
+        origin = (5, 6, 7)
+        cellsize = (2e-9, 3e-9, 4e-9)
+        origin_phys = np.array(origin) * np.array(cellsize)
+        world = World(cellsize)
+        magnet = Ferromagnet(world, Grid((11, 12, 13), origin=origin))
+        #set_parameters(magnet, 0)
+        #dw_profile(magnet, 0, 0)
+        assert np.all(np.isclose(world.window.position(), origin_phys))
```
ilateur: Careful with `np.isclose`. It has a default absolute tolerance of 1e-8, which is larger than the cellsize.

test/test_moving_window.py
```diff
@@ -0,0 +1,230 @@
...
+
+class TestShift:
```
ilateur, line 117: I think it would be nice to check that the average magnetization is still close to 0 every time as well, so you're not only checking whether or not it shifted at all, but that it shifted correctly.

test/test_moving_window.py
```diff
@@ -0,0 +1,230 @@
...
+        
+        world.center_domain_wall(comp, axis)
+        
+        magnet.jcur = (1e12, 0, 0)
+        world.timesolver.run(runtime)
+
+        left  = magnet.magnetization()[comp, :, :,  0].squeeze()
+        right = magnet.magnetization()[comp, :, :, -1].squeeze()
+
+        assert np.allclose((left_init, right_init), (left, right), atol=1e-4) # reduce atol when edge charges can be removed
```
ilateur: How will removing edge charges help if demag is already disabled?

ilateur (changes requested):
Sorry for the long review, but I hope it's useful and clear.

I have some general concerns as well:
- Edge charge compensation seems fairly important to a moving window. I think it should be within the same update, although not within the same pull request.
- I think there should be an example using the moving window, along with an explanation of the current restrictions (preferably on the website too).
- Don't forget to add the Window class to the documentation, via `docs/api.rst` I believe.
- `world.center_domain_wall` is now something you call once and it is then permanently enabled right? Should there be a way to disable this again? (It would also need something like `ClearPostSteps()` from mumax3 then.)

lamoreel:
> Edge charge compensation seems fairly important to a moving window. I think it should be within the same update, although not within the same pull request.

This PR is already in the making :blush: .

> I think there should be an example using the moving window, along with an explanation of the current restrictions (preferably on the website too).

Yes.

> Don't forget to add the Window class to the documentation, via `docs/api.rst` I believe.

It seems I keep forgetting this.

> `world.center_domain_wall` is now something you call once and it is then permanently enabled right? Should there be a way to disable this again? (It would also need something like `ClearPostSteps()` from mumax3 then.)

Very fair. On it!

src/core/window.hpp
lamoreel: Exactly.

src/core/window.cpp
lamoreel: `dir` only indicates the direction, so only its sign is relevant. If the wall is sufficiently fast to move more than the tolerance set in `shift`, the time step will decrease.

mumaxplus/world.py
lamoreel: There was, then I tried to include geometry support and it failed, so I removed it. Then I removed geometry support, so this can be included again :)

mumaxplus/world.py
lamoreel: In the mindset of backwards compatibility, I prefer not to. At one point (hopefully), this (and future) function(s) will be as general as it (they) can be.

src/physics/mumaxworld.cpp
lamoreel: If all sublattices are not sufficiently coupled, then the whole point of a moving window seems lost.

test/test_moving_window.py
lamoreel: Yes, I forgot to clean up :blush: .

mumaxplus/world.py
ilateur: Hooray!

test/test_moving_window.py
lamoreel: Exactly because I felt fancy when writing this.

src/core/window.hpp
lamoreel: `Window` is supposed to bridge `src/` and `physics/` and is kept very general as to support future functionalities such as 2D motion. I agree on the fact that `centerDomainWall` from `MumaxWorld` could perhaps be moved to `Window`, which might make more sense. Other than that, I prefer to keep it separate from `Grid` since it does contain (and needs) information about the relevant physics.

docs/images/moving_window.mp4
DiegoDeGusem, line 1: The x-axis on this animation does not change so it seems as if the DW is just jumping a bit back and forth and changing colour.

DiegoDeGusem (changes requested):

docs/examples/moving_window.rst
```diff
@@ -0,0 +1,22 @@
+:nosearch:
+
+Moving simulation window
+===============
```
ilateur: `===` should have as many characters as the title I think.

docs/examples/moving_window.rst
ilateur, line 1: This file is not used in the documentation. You need to add it to `examples.rst`.

docs/examples/moving_window.rst
```diff
@@ -0,0 +1,22 @@
...
+
+In this example we move a domain wall in a ferromagnet using a Zhang-Li STT. We let the simulation
+window move together with the wall, keeping the domain wall centered in the simulation space.
+Using this, we can virtually simulate an infitly long magnetic nanowire using a limited number
```
ilateur: *infinitely

docs/images/moving_window.mp4
ilateur, line 1: This is not so much a comment on the code, but more of a question about physical correctness.

In the video you can see the domain wall jump back a bit with each re-centering. You can also see this in the domain wall position.
Image
I find this slightly worrying, as this should be perfectly compensated for, or is this fine?

examples/moving_window.py
```diff
@@ -0,0 +1,115 @@
+"""
+In this example we move a domain wall in a ferromagnet using a Zhang-Li STT. We let the simulation
+window move together with the wall, keeping the domain wall centered in the simulation space.
+Using this, we can virtually simulate an infitly long magnetic nanowire using a limited number
```
ilateur: *infinitely

examples/moving_window.py
```diff
@@ -0,0 +1,115 @@
...
+magnet.magnetization = twodomain((1, 0, 0), (0, 1, 0), (-1, 0, 0), magnet.center[0], 5e-9)
+magnet.minimize()
+
+# ----------- Add current and simulate -----------
+magnet.jcur = (-1e12, 0, 0)
+magnet.xi   = 0.2
+magnet.pol = 1
+
+# Center the simulation window, keeping component 0 (x) close to zero.
+# We expect motion alongt he x axis.
```
ilateur: *along the

docs/images/moving_window.mp4
ilateur: Ah I think I see it. This is not actually plotting the domain wall position, but rather the moving window position, which can only be multiples of cellsize. Sometimes it just needs no jump twice in quick succession with this particular velocity. The actual domain wall position (moving window position + DW position within the window) should behave more smoothly I think.

examples/moving_window.py
```diff
@@ -0,0 +1,115 @@
...
+
+# ----------- Create movie -----------
+fig, axes = plt.subplots(2, 1, figsize=(10, 7))
+
+# Time trace subplot
+ax_trace = axes[0]
+lines = {}
+lines["pos"], = ax_trace.plot([], [], '-')
+
+ax_trace.set_title("Domain wall position")
```
ilateur: (Technically it is plotting the window position, not the domain wall position)

examples/moving_window.py
```diff
@@ -0,0 +1,115 @@
...
+magnet.xi   = 0.2
+magnet.pol = 1
+
+# Center the simulation window, keeping component 0 (x) close to zero.
+# We expect motion alongt he x axis.
+world.center_domain_wall(comp=0, axis=0)
+
+tmax = 2e-9
+timepoints = np.linspace(0, tmax, 100)
+outputquantities = {"mag": lambda: magnet.magnetization(),
```
ilateur: (I suppose `"mag": magnet.magnetization` without `lambda` and `()` works too, as it already behaves as a function)

DiegoDeGusem (approved):

docs/images/moving_window.mp4
ilateur: This is not the solution I had expected (I expected moving labels, not increasing labels), but I suppose this works too

examples/moving_window.py
```diff
@@ -0,0 +1,123 @@
...
+def update(frame):
+    # Update image
+    ax_image.clear()
+    plot_field(output["mag"][frame], ax=ax_image, arrow_size=8)
+    ax_image.set_xlabel("$x$ (nm)")
+    ax_image.set_ylabel("$y$ (nm)")
+
+    # update x-ticks
+    ax_image.set_xticks(ticks)
+    ax_image.set_xticklabels([f"{t + output["pos"][frame] * 1e9:.0f}" for t in ticks])
```
ilateur: I wonder if `plot_field` should be updated to take a moving window into consideration, but this can be a different pull request.

mumaxplus/window.py
```diff
@@ -7,20 +7,20 @@ class Window:
...
     def _check_boundary(self, boundary):
         if not isinstance(boundary, int) or boundary not in (0, 1):
             raise ValueError(f"Invalid boundary: {boundary}. Must be one of (0, 1).")
 
-
     def insert_magnetization(self, boundary, value):
         """
         Set magnetization value at a given boundary.
-        If none are given (or set to zero), the current edge value is used.
+        If unset (or set to zero), the current edge value is used.
```
ilateur: But `value` is not an optional argument, so there is no "unset"? Maybe by default `value=0`, or simply
> If set to zero, [...]

mumaxplus/world.py
```diff
@@ -395,4 +396,10 @@ def center_domain_wall(self, comp, axis=None):
                         + f" The {('x', 'y', 'z')[axis]}-direction is used as normal to the"
                         + " domain wall", UserWarning)
 
+        av = magnet.magnetization.average() if isinstance(magnet, Ferromagnet) else magnet.sub1.magnetization.average()
+        if np.abs(av[comp]) > 1e-2:
```
ilateur: Why 1e-2? I feel like this should depend on the number of cells and the cellsize 🤔 but not sure

mumaxplus/window.py
```diff
@@ -32,17 +32,24 @@ def insert_magnetization(self, boundary, value):
         self._check_boundary(boundary)
         self._impl.insert_magnetization(boundary, value)
 
+    def disable_motion(self):
```
ilateur, line 35: So to enable motion you need to call `world.center_domain_wall(comp, axis)`. To disable it again you need to call `world.window.disable_motion()`? Is this clear enough? 🤔

docs/images/moving_window.mp4
ilateur: Is there a built-in domain wall position property to take both window position and average magnetization into account? Or do you need to do this yourself?

mumaxplus/window.py
```diff
@@ -0,0 +1,55 @@
+class Window:
+    """Simulation window of the world.
+
+    Each world already has its own Window. This Window can be accessed through
+    the world.window property.
+
+    Windows should not be initialized by the end user.
```
JonathanMaes, line 7: If the user shouldn't initialize `Window`s, should they be called `_Window`?

mumaxplus/util/__init__.py
```diff
@@ -27,7 +27,7 @@
     "show_magnet_geometry",
     "show_field_3D",
     "show_regions",
-    # voronoi
+    # misc
     "VoronoiTessellator",
-    "MFM"
+    "MFM",
```
JonathanMaes: why add a comma?

mumaxplus/world.py
```diff
@@ -96,6 +100,11 @@ def get_ncafm(self, name):
             raise KeyError(f"No magnet named {name}")
         return NcAfm._from_impl(magnet_impl)
 
+    @property
+    def magnets(self) -> dict[str,Magnet]:
+        """Get a dictionairy of all magnet names."""
```
JonathanMaes: Typo: dictionary (also in other places)

docs/images/moving_window.mp4
lamoreel: Yes, this is in fact showing the window position, which shifts cell by cell. The wall position in function of time is smooth, which I checked using another custom (python) function. Should the latter become a built-in method?
Edit: I incorporated this custom function in the example as to show users the benefits of a Python interface once again :smile:

mumaxplus/window.py
lamoreel: Perhaps. Then we should rename `TimeSolver` and `DmiTensor` as well :thinking:

mumaxplus/window.py
lamoreel: I think this might be best since we will have other functions like `center_skyrmion` at some point in the future. Then there's only one global method to disable any centering. I"ll add appropriate references in the docstrings.

mumaxplus/window.py
ilateur: hmm nah?

examples/moving_window.py
```diff
@@ -55,30 +55,46 @@
 magnet.pol = 1
 
 # Center the simulation window, keeping component 0 (x) close to zero.
-# We expect motion alongt he x axis.
+# We expect motion alongt the x axis.
```
ilateur: *along

examples/moving_window.py
```diff
@@ -104,12 +119,20 @@ def update(frame):
...
+    ax_image.set_xticks(tick_positions[mask])
+    ax_image.set_xticklabels([f"{t:.0f}" for t in ticks[mask]])
     ax_image.set_title(f"$t$ = {output['time'][frame] * 1e9:.3f} ns")
 
     # Update time trace
-    lines["pos"].set_data(np.array(output["time"][:frame+1]) * 1e9, np.array(output["pos"][:frame+1]) * 1e9)
+    lines["window position"].set_data(np.array(output["time"][:frame+1]) * 1e9,
+                                      np.array(output["window position"][:frame+1]) * 1e9)
+    lines["wall position"].set_data(np.array(output["time"][:frame+1]) * 1e9,
+                                    np.array(output["wall position"][:frame+1] - p0) * 1e9)
```
ilateur, line 134: why `- p0`? 🤔

ilateur (approved):

examples/moving_window.py
lamoreel: So both curves start at the origin. Otherwise the wall position will have an off-set of `length/2`.

ilateur: Ya, but ain't that the way it is?

lamoreel: Sure, but I want to maximize prettiness in example videos :innocent:

JonathanMaes (approved):

PR #164, opened 2026-06-25
Add chiral magnetoelastic field and force with tests

ilateur:
Materials of the cubic point group 23 (or B20 compounds) can have an additional chiral magnetoelastic coupling, with the following energy density (https://doi.org/10.5445/IR/1000184834).
$\mathcal{E} = B_c \sum_{i, j, k} \epsilon_{ijk} \varepsilon_{ii} m_j^2$
This pull request adds the effective field and force density resulting from this energy density, including unit tests.

### DIFF
--- a/mumaxplus/ferromagnet.py
+++ b/mumaxplus/ferromagnet.py
@@ -653,7 +653,7 @@ class Ferromagnet(Magnet):
         
         See Also
         --------
-        B2
+        B2, B_chiral
         """
         return Parameter(self._impl.B1)
 
@@ -678,7 +678,7 @@ class Ferromagnet(Magnet):
         
         See Also
         --------
-        B1
+        B1, B_chiral
         """
         return Parameter(self._impl.B2)
 
@@ -696,6 +696,28 @@ class Ferromagnet(Magnet):
             warnings.warn("The second magnetoelastic coupling constant B2"
                           + " is set to a positive value, instead of negative (or zero)."
                           + " Make sure this is intentional!", UserWarning)
+
+    @property
+    def B_chiral(self) -> Parameter:
+        r"""Chiral magnetoelastic coupling constant (J/m³).
+
+        Notes
+        -----
+        Materials of the cubic point group 23 (or B20 compounds) can have an additional chiral magnetoelastic coupling, with the following energy density [1]_.
+
+        .. math:: \mathcal{E} = B_\text{chiral} \sum_{i, j, k} \epsilon_{ijk} \varepsilon_{ii} m_j^2
+
+        .. [1] L\ . Franke, “Elastic Coupling at Quantum Phase Transitions and in Chiral Magnets,” Das Karlsruher Institut für Technologie, Karlsruhe, 2025. doi: 10.5445/IR/1000184834.
+        
+        See Also
+        --------
+        B1, B2
+        """
+        return Parameter(self._impl.B_chiral)
+
+    @B_chiral.setter
+    def B_chiral(self, value):
+        self.B_chiral.set(value)
 
     # ----- POISSON SYSTEM ----------------------
 
@@ -1128,7 +1150,7 @@ class Ferromagnet(Magnet):
 
         See Also
         --------
-        B1, B2
+        B1, B2, B_chiral
         Magnet.strain_tensor, Magnet.rigid_norm_strain, Magnet.rigid_shear_strain
         magnetoelastic_force
         """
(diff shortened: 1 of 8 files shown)


mumaxplus/ferromagnet.py
```diff
@@ -697,6 +697,28 @@ def B2(self, value):
                           + " is set to a positive value, instead of negative (or zero)."
                           + " Make sure this is intentional!", UserWarning)
 
+    @property
+    def B_chiral(self) -> Parameter:
+        r"""Chiral magnetoelastic coupling constant (J/m³).
+
+        Notes
+        -----
+        Materials of the cubic point group 23 (or B20 compounds) can have an additional chiral magnetoelastic coupling, with the following energy density [1]_.
```
DiegoDeGusem: Additional to what? I find this clearer
> Materials of the cubic point group 23 (or B20 compounds) can have a chiral magnetoelastic coupling, with the following energy density [1]_.

or something like

> Additionally to B1 and B2 magnetoelastic coupling, materials of the cubic point group 23 (or B20 compounds) can have a chiral magnetoelastic coupling, with the following energy density [1]_.

DiegoDeGusem (changes requested):

mumaxplus/ferromagnet.py
ilateur: I went with the first option, as `B1` and `B2` are already mentioned in `See Also`.

DiegoDeGusem (approved):

test/test_magnetoelasticfield.py
```diff
@@ -163,12 +192,10 @@ def test_random():
     magnet.enable_elastodynamics = True
 
     magnet.msat = msat
-    B1, B2 = B, 0.5*B
+    B1, B2, Bc = B, 0.5*B, 0.3*B
```
lamoreel, line 195: I'm guessing the sign of `B_chiral` also contains relevant physics, no? This can be tested here.

mumaxplus/ferromagnet.py
```diff
@@ -697,6 +697,28 @@ def B2(self, value):
                           + " is set to a positive value, instead of negative (or zero)."
                           + " Make sure this is intentional!", UserWarning)
 
+    @property
+    def B_chiral(self) -> Parameter:
```
lamoreel, line 701: Is `B_chiral` really an independent material parameter? Or is it linked with `B1` and `B2`? The different notation compared to the mentioned ref makes it hard to figure out.

ilateur: It creates an independent effect, linking strain and magnetization in a different way not achievable by any combination of `B1` and `B2`. Of course changing your material in order to change `B1` and `B2` will probably also change `B_chiral`.
In the reference these are
$\lambda_{11} = B_1, \qquad \lambda_{44} = B_2 / 2, \qquad \lambda_{12} = B_{\text{chiral}}$
with $\lambda_{11}$, $\lambda_{44}$ and $\lambda_{12}$ as in equation (8.16).

Is it useful to reference specific equations or symbols from that source in the documentation?

test/test_magnetoelasticfield.py
ilateur: Despite the chiral name, nothing special will happen to the magnetic field. It will just change sign.

mumaxplus/ferromagnet.py
lamoreel: Perhaps, yes (at least one person (myself) was unable to make a one-on-one correspondence right away :blush: ). But it's your call!

jplauzie: Personally I tend to find having the equations worked out useful, for what it is worth. If you include equations, for those particular symbols, I would maybe emphasize somewhere that the λ's are not the normal magnetostrictive coefficients though

mumaxplus/ferromagnet.py
```diff
@@ -703,10 +703,16 @@ def B_chiral(self) -> Parameter:
 
         Notes
         -----
-        Materials of the cubic point group 23 (or B20 compounds) can have a chiral magnetoelastic coupling, with the following energy density [1]_.
+        Materials of the cubic point group 23 (or B20 compounds) can have a
+        chiral magnetoelastic coupling, with the following energy density.
 
         .. math:: \mathcal{E} = B_\text{chiral} \sum_{i, j, k} \epsilon_{ijk} \varepsilon_{ii} m_j^2
```
lamoreel, line 709: Maybe also specify that `\varepsilon` denotes the strain, as to clearly distinguish with `\epsilon`.

lamoreel (approved):

PR #171, opened 2026-08-05
Add Magnetoelastics to the minimizer

jplauzie:
Hi,

This is a first pass at extending the steepest descent minimizer to work with elastics/magnetoelastics. The main changes are in minimizer.cu/removerigidbodies.cu. This shouldn't change any existing behavior if magnetoelastics are not enabled.

This takes the same approach that the existing magnetic steepest descent minimizer does, but for the elastic system: Do a steepest descent method on the displacement using just the force, dropping the transient eta/velocity term (as the precession term gets dropped in magnetics), with Barzilian-Borwein steps. The characteristic step sizes for the magnetic and elastic systems are very different, so magnetic and elastic steps are alternated until both systems hit their separate convergence criteria. 

There are two big differences compared to the magnetic minimizer: 
1) The dudu term in the BB1 rule has a habit of underflowing in float32 while it's being constructed, so there is a temporary rescaling, but this falls out once the actual BB step is constructed and has no effect on the results. 

2) to match the magnetic minimizer, the convergence criteria is a lastDu (one could instead monitor the force?). Rigid body modes (uniform translation/rotation) will inflate this, causing the minimizer to take many more steps than is necessary, even after it's reached equilibrium, so they're just subtracted off. Currently, I subtracted them off of u/force directly. You could maybe do this just for the convergence check, instead of u itself. Convergence wise, either way is fine. I noticed the values for u can get quite large in long runs/minimizes, ~3e-9 or so, and I wasn't sure if that could cause noise problems if du is say ~1e-16 or so in float32, so i went with cleaning u directly. There is a flag to disable removerigidbodies, just in case someone doesn't want it.

I think minimizer.cu is mostly ready to go, absent a few minor tweaks. I need to pull out the diagnostic messages, and probably removerigidbodies does not need to be done every step, so performance-wise every 100steps or so is fine.  The minimizer portion will work without removerigidbodymodes, it just inflates the number of steps it takes if it's not used.
 
removerigidbodymodes works mechanically (at least for simple test cases of uniform rho and a plain 2-D geometry). There's definitely some nonsense in terms of readability, unnecessary use of doubles on the GPU, etc. In principle it has the code to handle non-uniform rho/other geometries, but I've only tested against uniform 2-D sheets after i ran into the donut issue. I'm back from visiting family, so I should be able to get this cleaned up in the next few days.
 
So far, I've only tested it with FMs, with a uniform 2-D geometry, and a few starting magnetic configurations (uniform, vortex, etc), and it converges to the same states that a long run (25ns) does. It alsohandles a user-applied body-force.  The function hooks are there for it to work with AFMs/altermagnets, but I don't know as much about them for magnetoelastics so I haven't tested them yet. I also don't know what would be a good test script for the test suite?

This PR also has 1 extra feature; I added a call to clean rigid modes to the time solver. This is still experimental and needs to be tested further, but it seems to be allowing the adaptive time solver to take larger steps for some scripts (not all). I'm still confirming it's not spurious, but I did not expect it to do anything at all, so that was a bit surprising.

[magnetoelastictest3.txt](https://github.com/user-attachments/files/30744759/magnetoelastictest3.txt)

Attached is a script which shows a significant difference (on my machine, with cleaned rigid modes it takes ~2:15 to run. Without, ~9:52). So far, it seems to be scripts where PBC is disabled (so the no-traction BCs are used), and that are run for at least a few ns where there is any difference. My initial guess would be, cleaning rigid modes is compensating for some of the instability generated by the no-traction BCs that would otherwise build up over time? (Alas, not all of it. The donut shaped geometry is still unstable). More typical scripts like the ME dispersion script show no difference.


ilateur:
Thank you very much for your contributions Josh!

Your last point of speeding up regular `run` simulations by potentially removing instabilities is very interesting :thinking:

jplauzie:
Hi,

Sorry the clean up update took longer than I expected. 😅

I would say minimizer.cu is now 'done', barring any comments. rigidbodymodes.cu is mostly done, and should be readable now. I still need to check 2 small things: On extra usage of unnecessary doubles in the cuda kernels, and also I can probably optimize the reduction kernels a bit more using warp shuffles.

I took some time to look more into other methods to see how people handle rigid body modes, and I basically came across 5 methods, each with pros&cons( short version: I kept 1 and 4):

1) linearized removal of translations/rotations, which is what rigidbodymodes already does

2) projecting the modes out using a Gram-Schmidt process, as [Petsc](https://petsc.org/release/manualpages/Mat/MatNullSpaceCreateRigidBody/) does. Also linearized.

3) An SVD process (Kabsch/Procrustes method)

4) Quaternions (Horn's method)

5) Fixing a few cells. This is what [Comsol](https://doc.comsol.com/6.4/doc/com.comsol.help.sme/sme_ug_solid.07.069.html) uses.

1&2 both work to clean displacement u and force f (and v, for dynamics, but not necessary in the minimizer), but are linearized, so they're assuming small rotations. 3&4 can handle arbitrary large rotations, but I wasn't able to get them to apply to forces (and couldn't find any examples of them being used for that), only displacements. 5 would require porting over the FixDispLoc/FixDispVal features from mumax3 ME which don't exist yet in mumaxplus, and also requires the user to very careful in picking the fixed points (otherwise it can introduce artificial stresses/strains). Because of that, I did not look into 5 further.

In my quick demo test, 1&2 gave identical results to each other but 1 was noticeably faster, so I stuck with 1. This is the one used every step to clean u and f in the minimizer. However, they do assume a small angle approximation. In general, this seems to be fine, in all the test cases I tried the angle was small, even after a relatively long (5ns+) run. I think realistically, this only might happen if a user calls minimize after a run(), where that run was applying a user-body-force with a significant rigid rotational component? To handle that, I included method 4, which can be called once at the start of a minimize, to get rid of an initial large rotation in the displacement. (3&4 were giving identical results to each other. I went with 4 since it seems slightly more robust but kind of arbitrary). 

If there is interest in 2 and 3, I wouldn't mind cleaning them up and adding them, but I didn't see a point adding redundant functions unless there is interest. 

This update adds Method 4, which is functional (it gives the same angle value as method 1 for small angles), but I haven't gotten to cleaning it up yet, I wouldn't waste time reviewing it yet, there be dragons. I will do that after I finish the 2 things left in rigidbodymodes1.  I can also split it off into it's own PR if that ends up being easier.

ilateur:
Good stuff!

> If there is interest in 2 and 3, I wouldn't mind cleaning them up and adding them, but I didn't see a point adding redundant functions unless there is interest.

To me there is no point in adding redundant functions, unless someone would be interested in specifically comparing these methods.

Also recently #161 added the `postStep()` function to the timesolver. It might be nice to add one of these functions here (nr 1 by the sound of it) :thinking:. Perhaps with the option of disabling it but enabled by default? I'm not sure how clean and user-friendly this `postStep` interface is right now, as it was specifically implemented for moving windows. (This can be a different pull request and you don't have to volunteer for this, just writing down the idea.)

lamoreel:
> I'm not sure how clean and user-friendly this `postStep` interface is right now, as it was specifically implemented for moving windows. (This can be a different pull request and you don't have to volunteer for this, just writing down the idea.)

`postStep()` is called after every timesolver step and holds on to a pointer to the actual function (which you can set and unset), so this should be pretty general. However, combining multiple functions will be messy this way and can probably be improved.

src/physics/rigidbodymodes_common.hpp
```diff
@@ -0,0 +1,66 @@
+#pragma once
+
+#include "datatypes.hpp"
+#include "cudalaunch.hpp"
```
DiegoDeGusem: Hi, thanks for the work! However, when I try compiling this line gives an error. I think you can only include `cudalaunch.hpp` in `.cu` files. I solved this problem by creating a new file called `rigidbodymodes_common.cu` which looks like this
```c++
#include "cudalaunch.hpp"
#include "rigidbodymodes_common.hpp"

// T is deduced from the sdata array's element type
template <int NFIELDS, typename T>
__device__ inline void blockReduceSum(T sdata[NFIELDS][BLOCKDIM], int tid) {
  for (unsigned int s = BLOCKDIM / 2; s > 0; s >>= 1) {
    if (tid < s) {
      #pragma unroll
      for (int f = 0; f < NFIELDS; f++)
        sdata[f][tid] += sdata[f][tid + s];
    }
    __syncthreads();
  }
}
```
In `rigidbodymodes_common.hpp` I then removed the `#include cudalaunch.hpp` and changed the funcion that is now in the `.cu` file to
```C++
// T is deduced from the sdata array's element type
template <int NFIELDS, typename T>
__device__ inline void blockReduceSum(T sdata, int tid) {}
```
Now I can compile without compilation errors.

DiegoDeGusem (changes requested):

src/physics/rigidbodymodes_common.hpp
JonathanMaes: On Windows, I get no compilation errors with the original code, and all tests pass.
However, I do get two new warnings during compilation:
```pwsh
src\physics\rigidbodymodes_common.hpp(48,15): warning C4068: unknown pragma 'unroll' [build\temp.win-amd64-cpython-313\Release\single\src\physics\physics.vcxproj]
src\physics\rigidbodymodes_common.hpp(62,11): warning C4068: unknown pragma 'unroll' [build\temp.win-amd64-cpython-313\Release\single\src\physics\physics.vcxproj]
```

When appyling Diego's suggested fix, only the second warning remains. It seems, therefore, that `#pragma unroll` is only supported in .cu files (at least for me). It would be nice if the one in `warpReduceSum` could also be moved.

jplauzie: I'm on Windows as well. I didn't notice the warnings previously, but I was getting them as well. I've moved both as suggested.

Should `rigidbodymodes_common.cu` be `rigidbodymodes_common.cuh` ? I don't know if there's a functional difference, but i think .cuh is Cuda's equivalent to .h or .hpp. I don't see it commonly used

mumaxplus/altermagnet.py
```diff
@@ -133,12 +140,21 @@ def minimize(self, tol=1e-6, nsamples=20):
...
             The number of consecutive magnetization evaluations that must not
             differ by more than the tolerance "tol".
+        tol_el : int / float (default=1e-6)
+            The maximum allowed difference between consecutive elastic
+            displacement evaluations when advancing toward mechanical
+            equilibrium. Ignored if elastodynamics is disabled.
+        nsamples_el : int (default=10)
+            The number of consecutive elastic displacement evaluations that
+            must not differ by more than the tolerance "tol_el". Ignored if
+            elastodynamics is disabled.
```
JonathanMaes, line 150: A short description of the other new arguments (which can be copied from `ncafm.py`) may be useful as well (same comment for `antiferromagnet.py`, `ferromagnet.py` and `world.py`).

Or just describe those technical arguments once in `world.py` and refer there in all the other minimize docstrings.

mumaxplus/antiferromagnet.py
```diff
@@ -117,27 +117,42 @@ def enable_demag(self, value):
...
+        check against `tol_el`/`nsamples_el`.
+
         Parameters
         ----------
         tol : int / float (default=1e-6)
             The maximum allowed difference between consecutive magnetization
             evaluations when advancing toward an energy minimum.
-
-        nsamples : int (default=20)
+        nsamples : int (default=10)
```
JonathanMaes: Default value is 20.

mumaxplus/antiferromagnet.py
```diff
@@ -117,27 +117,42 @@ def enable_demag(self, value):
...
             The number of consecutive magnetization evaluations that must not
             differ by more than the tolerance "tol".
+        tol_el : int / float (default=1e-6)
+            The maximum allowed difference between consecutive elastic
+            displacement evaluations when advancing toward mechanical
+            equilibrium. Ignored if elastodynamics is disabled.
+        nsamples_el : int (default=20)
```
JonathanMaes, line 145: Default value is 10.

mumaxplus/ncafm.py
```diff
@@ -125,27 +125,57 @@ def enable_demag(self, value):
...
+        check against `tol_el`/`nsamples_el`.
 
         Parameters
         ----------
         tol : int / float (default=1e-6)
             The maximum allowed difference between consecutive magnetization
             evaluations when advancing toward an energy minimum.
-
-        nsamples : int (default=30)
+        nsamples : int (default=10)
```
JonathanMaes: Default value is 30 (also for `nsamples_el`).

mumaxplus/ncafm.py
```diff
@@ -125,27 +125,57 @@ def enable_demag(self, value):
...
+            regardless of convergence.
+        rigid_body_modes_interval : int (default=1)
+            How often (in steps) rigid-body translation and rotation modes
+            are projected out of the elastic displacement and force. A
+            value of 1 removes them every step, n removes them every n
+            steps, and 0 disables removal entirely. 
+        rigid_body_modes_delay : int (default=0)
+            The number of steps to wait before rigid-body-mode removal
+            begins, regardless of "rigid_body_modes_interval". 
+
```
JonathanMaes, line 172: `rigid_body_modes_method` description is missing.

mumaxplus/ncafm.py
```diff
@@ -125,27 +125,57 @@ def enable_demag(self, value):
...
             The number of consecutive magnetization evaluations that must not
             differ by more than the tolerance "tol".
+        tol_el : int / float (default=1e-6)
+            The maximum allowed difference between consecutive elastic
+            displacement evaluations when advancing toward mechanical
+            equilibrium. Ignored if elastodynamics is disabled.
+        nsamples_el : int (default=10)
+            The number of consecutive elastic displacement evaluations that
+            must not differ by more than the tolerance "tol_el". 
+        stepsize_el : int / float (default=1e-30)
```
JonathanMaes, line 155: Is a default value of 1e-30 correct here? In the other minimizes it is 1e-14.

mumaxplus/world.py
```diff
@@ -347,63 +347,4 @@ def unset_pbc(self):
...
-
-        av = magnet.magnetization.average() if isinstance(magnet, Ferromagnet) else magnet.sub1.magnetization.average()
-        if np.abs(av[comp]) > 4 / magnet.grid.size[0]:
-            raise RuntimeError(f"The domain wall does not seem centered (average {('x', 'y', 'z')[comp]}-"
-                              + f"component is {av[comp]:.2e}). `center_domain_wall` only works properly "
-                              + "if the wall is initialized near the center of the magnet.")
-
-        self._impl.center_domain_wall(comp, axis)
\ No newline at end of file
+        self._impl.unset_pbc()
```
JonathanMaes: Newline at end of file disappeared.

(Same comment for `physics/altermagnet.cpp`, `antiferromagnet.cpp`, `ferromagnet.cpp`, `minimizer.cu`, `minimizer.hpp`, `mumaxworld.cpp`, `mumaxworld.hpp`, `rigidbodymodes4.cu` and `rigidbodymodes4.hpp`.)

src/bindings/wrap_altermagnet.cpp
```diff
@@ -37,7 +37,14 @@ void wrap_altermagnet(py::module& m) {
       .def_readonly("dmi_tensor", &Altermagnet::dmiTensor)
       .def_readonly("dmi_vector", &Altermagnet::dmiVector)
 
-      .def("minimize", &Altermagnet::minimize, py::arg("tol"), py::arg("nsamples"))
+      .def("minimize", &Altermagnet::minimize,
+           py::arg("tol"), py::arg("nsamples"),
+           py::arg("tol_el") = 1e-6, py::arg("nsamples_el") = 20,
+           py::arg("stepsize_el") = 1e-30, py::arg("stepsize_el_fallback") = 1e-30,
```
JonathanMaes: The default values of some of these (`nsamples_el` and `stepsize_el`) do not correspond to their definitions in the `.py` files.

Also, `tol` and `nsamples` don't (and didn't) have their default values listed here, so I don't think it is necessary to declare all the default values here?

(Same comment for `wrap_antiferromagnet.cpp`, `wrap_ferromagnet.cpp`, `wrap_ncafm.cpp` and `wrap_world.cpp`.)

src/bindings/wrap_world.cpp
```diff
@@ -146,10 +151,5 @@ void wrap_world(py::module& m) {
...
       .def_property("pbc_repetitions", &MumaxWorld::pbcRepetitions,
                     &MumaxWorld::setPbcRepetitions, "PBC repetitions of the world")
-
-      // Moving simulation window
-      .def_property_readonly("window", &MumaxWorld::window,
-                             py::return_value_policy::reference)
-      .def("center_domain_wall", &MumaxWorld::centerDomainWall, py::arg("comp"), py::arg("axis"))
      ;
-}
+}
```
JonathanMaes: Newline at end of file disappeared.

src/physics/ncafm.hpp
```diff
@@ -34,8 +34,12 @@ class NcAfm : public HostMagnet {
  const Ferromagnet* sub2() const;
  const Ferromagnet* sub3() const;
 
- void minimize(real tol = 1e-6, int nsamples = 30);
- void relax(real tol);
+void minimize(real tol = 1e-6, int nSamples = 30,
+              real tolEl = 1e-6, int nSamplesEl = 10,
```
JonathanMaes: nSamplesEl is default 30 in the Python layer

src/bindings/wrap_magnet.cpp
```diff
@@ -31,6 +31,12 @@ void wrap_magnet(py::module& m) {
       .def_property("enable_elastodynamics",
                     &Magnet::enableElastodynamics,
                     &Magnet::setEnableElastodynamics)
+      .def_property("enable_elastodynamics",
```
JonathanMaes: enable_elastodynamics is now defined twice?

src/physics/altermagnet.hpp
```diff
@@ -34,8 +34,12 @@ class Altermagnet : public HostMagnet {
  const Ferromagnet* sub1() const;
  const Ferromagnet* sub2() const;
  
- void minimize(real tol = 1e-6, int nSamples = 20);
- void relax(real tol);
+void minimize(real tol = 1e-6, int nSamples = 20,
+              real tolEl = 1e-6, int nSamplesEl = 10,
+              real stepsizeEl = 1e-30, real stepsizeElFallback = 1e-30,
```
JonathanMaes, line 39: `stepsizeEl` is 1e-14 in the Python layer. Same goes for `antiferromagnet.hpp`.

src/physics/rigidbodymodes4.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+class Field;
+class Magnet;
+
+// ---------------------------------------------------------------------------
+// Exact (non-linearized) rigid-body alignment via Horn's closed-form
+// quaternion method (Horn, 1987, "Closed-form solution of absolute
+// orientation using unit quaternions").
+//
+// Mathematically equivalent to the Kabsch/SVD approach in rigidbodymodes3.cu
```
JonathanMaes: What is rigidbodymodes3.cu?

JonathanMaes (changes requested):
I noticed some small inconsistencies in the default values between magnets.
I have not yet looked at the files with >100 lines changed, but to avoid waiting too long I'll already post the comments I have now. Either way, everything works on my machine without any warnings and all tests pass successfully.

src/bindings/wrap_altermagnet.cpp
jplauzie: Oops, looks like a double bug here, on my part. Not having the default values listed was causing a compile error, but that was because the values in the `.py` files and `.hpp` files didn't match. Now that they match, these defaults are not needed

jplauzie:
ok, I am happy with rigidbodymodes.cu, I would say it's 'done' and ready to be looked over. I will start cleaning up rigidbodymodes4 now.

---

Given the discussion in #175 and [415](https://github.com/mumax/3/pull/415) , I guess I should probably explain a bit the logic of the reduction kernels, since they are slightly different from the existing ones, and you guys can decide if parts of it are worth keeping as is or are overkill. Happy to revert any parts that seem unnecessary:

The base is the same reduction kernel, based on the parallel reduction method outlined by Mark Harris and used in the existing sum kernels in mumax3/mumax+. I then converted it to use warp shuffles (introduced with CC 3.0), following this [tutorial](https://docs.olcf.ornl.gov/training/training_archive.html) by Bob Crovella at Nvidia, the 'CUDA Atomics Reductions and Warp Shuffle)' session. Warpshuffles are less instructions, less pressure on shared memory, and they require less synchronization than the traditional shared memory method, so they should be overall faster/more efficient.

For the number of blocks, since it's so problem-specific, I couldn't find an official source, but that sort of choice is listed in things like [Nvidia's programming guide](https://docs.nvidia.com/cuda/pdf/CUDA_C_Programming_Guide.pdf). If you guys want, we can just limit to number of SMs, like in [415](https://github.com/mumax/3/pull/415) ? I did notice a performance difference, ~85 GB/s vs 190 GB/s (maybe avoidable if things are refactored?)

The last two changes are to the grid-stride loop accumulations, and the accumulation at the end. When you sum up many floating point numbers, for large sums you can start to get an accumulation error that scales between $(n)^{1/2}-n$. This shouldn't matter for small grids, but maybe matters for large grids depending on how many blocks you have. For instance, if you have a 2048x2048x1 simulation, 48 blocks, with 256 threads/block(12,288 threads), each accumulator should sum up ~341 terms). 

There are different ways to deal with it (Kahan summation, maintain more accumulators/do more tree reductions, operate in double, etc). For the grid-stride accumulation, i went with [Kahan summation](https://en.wikipedia.org/wiki/Kahan_summation_algorithm). It requires a few extra FLOPs, and some extra register pressure, but in my testing since the kernel is bandwidth bound it was actually just free, no change in timings. It is a way to preserve extra digits of accuracy without the same performance penalty as using float64 operations.

For the final accumulation, i went with a second kernel launch that sums nblocks, in double. My original implementation actually summed it on the CPU, just like [415](https://github.com/mumax/3/pull/415), but I had issues getting it to play nice with the templating. It also seems ever so slightly more sensitive to the symmetry breaking applied field in the very first step, which is a good thing to maintain if possible (but this could be handled in other ways). I believe it maintains the determinism concern as well? Because this final accumulation is so small, the performance difference is negligible, like ~1%.

They do seem to have an effect on the number of steps, but I'm not sure how much of that is because it is more precise vs the minimizer taking a different path.

I don't have a strong opinion on which method, but it seems worth having some version to reduce the accumulation errors, especially given they're basically ~free.

edit:
I should add, there are 2 potential further optimizations:
Using the __restrict__ keyword makes the kernels ~5% faster, but required tweaking other files (I think field.hpp/parameter.hpp). I wasn't sure what broader impact that might have so I left this out. Nvidia also has the CUB library which is supposed to be faster than a handbuilt kernel, but I don't have any experience in how to apply

jplauzie:
I am now happy with rigidbodymodes4.cu (now renamed to rigidbodymodesQ.cu). With that, all 3 of the main files and their associated headers should be presentable/reviewable. I just need to finish resolving the small stuff mentioned in comments above

src/physics/rigidbodymodes4.hpp
jplauzie: Oops, that is a left over stale reference. Originally I had prototyped the first 4 methods mentioned [here](https://github.com/mumax/plus/pull/171#issuecomment-5380643100), and put each method in a separate rigidbodymodes1-4 file ( rigidbodymodes3.cu was the SVD process (Kabsch/Procrustes method)). It doesn't have any effect on what I committed, it was a completely isolated implementation and I eventually stripped out methods 2 and 3 to reduce the commit size since they were redundant.

mumaxplus/ncafm.py
jplauzie: 1e-30 is the correct stepsize for the elastics portion, 1e-14 should be maintained for the magnetic portions. They end up being at different scales/units for displacement/forces compared to magnetization/torques (Atleast for CoFeB material parameters, the BB stepsizes are around 1e-27 to 1e-31 or so. That might be a different range for other material parameters).

To handle the different step sizes, currently the magnetic and elastic system are both stepped separately, and it just alternates between the two until they both satisfy their convergence criteria. They effectively only see each other in that the magnetic system seems a ME torque term, and the elastic system sees a ME body force. In principle, there is probably a way to unify them into a single step as discussed by @ilateur [here](https://github.com/mumax/plus/pull/174#issuecomment-5424274526). (Some kind of rescaling/nondimensionalizing, a generalized gradient, etc?). But so far I haven't figured out a way to do it. In addition to the units, the other weird part is u/du does not have a natural scale, whereas m is always O(1).

This also shows up in the tolerances. Tol_el is nominally 1e-6, but this is actually a relative tolerance, unlike tol for the magnetic portion.

src/physics/rigidbodymodesQ.cu
DiegoDeGusem, line 1: When I try to build this branch (on Linux) I get the following error:
```bash
rigidbodymodesQ.cu(228): error: namespace "std" has no member "numeric_limits"
    const double EPS = std::numeric_limits<double>::epsilon();
                            ^
```
I could solve this by adding `#include <limits>` next to the other includes in this file (`rigidbodymodesQ.cu`).

DiegoDeGusem (changes requested):
Sorry in advance for this small change requests. I have another problem when building the code (see below) and I just want to make sure it can be built before I check the contents. 😄

src/physics/minimizer.cu
```diff
@@ -1,78 +1,213 @@
...
+    while (!converged())
+      step();
+    return;
+  }
+
+
+  while (!converged() && nsteps_ < maxSteps_) {
     step();
+  }
+  std::cerr << "steps: " << nsteps_ << " ." << std::endl;
```
DiegoDeGusem, line 210: I think I would remove this print statement in the final version?

src/physics/minimizer.cu
```diff
@@ -138,22 +289,162 @@ void Minimizer::step() {
 
     addMagDiff(maxVecNorm(dm));
   }
+}
+
+
+void Minimizer::stepElastic() {
+  for (size_t i = 0; i < elMagnets_.size(); i++) {
+    u0[i] = elMagnets_[i]->elasticDisplacement()->eval();
+    //if (shouldRemoveRigidBodyModes()) removeRigidBodyModes(u0[i], rigidGeoms_[i], elMagnets_[i],true);
```
DiegoDeGusem, line 298: Should this commented if statements be removed?

src/physics/minimizer.cu
```diff
@@ -138,22 +289,162 @@ void Minimizer::step() {
...
+
+    if (nsteps_ == 0) {
+      elMagnets_[i]->elasticDisplacement()->set(u0[i]);
```
DiegoDeGusem, line 301: I think this currently does nothing.

src/physics/minimizer.cu
```diff
@@ -138,22 +289,162 @@ void Minimizer::step() {
 
     addMagDiff(maxVecNorm(dm));
   }
+}
+
+
+void Minimizer::stepElastic() {
```
DiegoDeGusem: All for loops in this function are the same (`for (size_t i = 0; i < elMagnets_.size(); i++)`), so I was wondering if it would be possible to make it just one for loop.

src/physics/minimizer.cu
```diff
@@ -138,22 +289,162 @@ void Minimizer::step() {
...
+        elStepsizes_[i] = newHstep;
+      } else {
+        elStepsizes_[i] = stepsizeElFallback_;
+      }
+    } else {
+      elStepsizes_[i] = stepsizeElFallback_;
+    }
+
+    real maxU = maxVecNorm(u1[i]);
+    real relDu = (maxU > 0) ? (maxDu / maxU) : maxDu;
```
DiegoDeGusem, line 388: I would put this in another function analogous to `BarzilianBorweinStepSize` for the magnetic step size.

src/physics/minimizer.cu
```diff
@@ -138,22 +289,162 @@ void Minimizer::step() {
...
+    //if (shouldRemoveRigidBodyModes()) removeRigidBodyModes(u0[i], rigidGeoms_[i], elMagnets_[i],true);
+
+    if (nsteps_ == 0) {
+      elMagnets_[i]->elasticDisplacement()->set(u0[i]);
+      f0[i] = forces_[i].eval();
+    } else {
+      f0[i] = f1[i];
+    }
+
+    //if (shouldRemoveRigidBodyModes()) removeRigidBodyModes(f0[i], rigidGeoms_[i], elMagnets_[i],true);
```
DiegoDeGusem, line 307: Another commented if statement

src/physics/minimizer.cu
```diff
@@ -138,22 +289,162 @@ void Minimizer::step() {
...
+    cudaLaunch(ncells, k_stepElastic, u1[i].cu(), u0[i].cu(), f0[i].cu(), h);
+  }
+
+  for (size_t i = 0; i < elMagnets_.size(); i++) {
+   if (shouldRemoveRigidBodyModes()) removeRigidBodyModes(u1[i], rigidGeoms_[i], elMagnets_[i],true);
+  }
+
+  for (size_t i = 0; i < elMagnets_.size(); i++) {
+    elMagnets_[i]->elasticDisplacement()->set(u1[i]);
+  }
```
DiegoDeGusem, line 320: Can't these be combined in 1 for loop? Because I don't think that removing the body modes of one magnet influences displacement of other magnets.

src/physics/minimizer.cu
```diff
@@ -138,22 +289,162 @@ void Minimizer::step() {
...
+
+  for (size_t i = 0; i < elMagnets_.size(); i++) {
+    f1[i] = forces_[i].eval();
+    if (shouldRemoveRigidBodyModes()) removeRigidBodyModes(f1[i], rigidGeoms_[i], elMagnets_[i],true);
+  }
+
+  for (size_t i = 0; i < elMagnets_.size(); i++) {
```
DiegoDeGusem, line 327: I also think these 2 for loops can be combined, because at first glance you only need f1[i] and the others don't matter.

DiegoDeGusem (changes requested):
I went over the code and also tested it with something small. I initialized a ferromagnet with magnetization in the +x direction and DMI (the DMI should change the magnetization at the edges slightly which should then bend the magnet).
I then relaxed the magnetization and minimized the system which resulted in the magnetization, displacement and strain tensor shown below. These indeed show a slight bending of the magnet (🎉). 
Finally I let the system run a while (without external inputs) and I could not see any elastic/magnetic changes. This seems like a good indication that the system reached a minimized state. 😄 🎉

Image

Image

Image

PR #172, opened 2026-08-13
Make overlap more flexible

DiegoDeGusem:
With this PR it should now be possible to have overlapping grids without overlapping geometries.

The function is added to `MumaxWorld` because Grid has no access to geometries. It still uses the old `Grid::overlaps` function as a quick first check to see if the grids overlap. If they do overlap it uses a GPU function in `reduce.cu` to check if geometry cells overlap.

Two tests were also added to `test_geometry`, one where the geometries don't overlap and one where they do.

### DIFF
--- a/src/core/reduce.cu
+++ b/src/core/reduce.cu
@@ -290,3 +290,55 @@ bool isUniformField(const Field& f) {
   }
   return true;
 }
+
+__global__ void k_geometriesOverlap(bool* result,
+                                     Grid grid1, bool const* geometry1,
+                                     Grid grid2, bool const* geometry2,
+                                     int3 lo, int3 n, int ncells) {
+  __shared__ bool sdata[BLOCKDIM];
+  int tid = threadIdx.x;
+
+  bool found = false;
+  for (int i = tid; i < ncells; i += BLOCKDIM) {
+    // unravel flat index i into a 3D offset within [0, n)
+    int x = i % n.x;
+    int y = (i / n.x) % n.y;
+    int z = i / (n.x * n.y);
+    int3 coo{lo.x + x, lo.y + y, lo.z + z};
+
+    bool in1 = !geometry1 || geometry1[grid1.coord2index(coo)];
+    bool in2 = !geometry2 || geometry2[grid2.coord2index(coo)];
+    if (in1 && in2) {
+      found = true;
+      break;  // Once one overlap is found we can stop
+    }
+  }
+
+  sdata[tid] = found;
+  __syncthreads();
+
+  // Reduce the block
+  for (unsigned int s = BLOCKDIM / 2; s > 0; s >>= 1) {
+    if (tid < s)
+      sdata[tid] = sdata[tid] || sdata[tid + s];
+    __syncthreads();
+  }
+
+  if (tid == 0)
+    *result = sdata[0];
+}
+
+bool geometriesOverlap(Grid grid1, bool const* geometry1,
+                       Grid grid2, bool const* geometry2,
+                       int3 lo, int3 n) {
+  int ncells = n.x * n.y * n.z;
+
+  GpuBuffer<bool> d_result(1);
+  cudaLaunchReductionKernel(k_geometriesOverlap, d_result.get(),
+                            grid1, geometry1, grid2, geometry2, lo, n, ncells);
+
+  bool result;
+  checkCudaError(cudaMemcpyAsync(&result, d_result.get(), sizeof(bool),
+                                 cudaMemcpyDeviceToHost, getCudaStream()));
+  return result;
+}
--- a/src/core/reduce.hpp
+++ b/src/core/reduce.hpp
@@ -5,6 +5,7 @@
 #include "datatypes.hpp"
 
 class Field;
+class Grid;
 
 real maxAbsValue(const Field&);
 real maxVecNorm(const Field&);
@@ -14,3 +15,6 @@ std::vector<real> fieldAverage(const Field&);
 bool idxInRegions(GpuBuffer<unsigned int>, unsigned int idx);
 bool isUniformFieldComponent(const Field&, int);
 bool isUniformField(const Field&);
+bool geometriesOverlap(Grid grid1, bool const* geometry1,
+                       Grid grid2, bool const* geometry2,
+                       int3 lo, int3 n);
--- a/src/physics/mumaxworld.cpp
+++ b/src/physics/mumaxworld.cpp
@@ -15,6 +15,7 @@
 #include "magnet.hpp"
 #include "minimizer.hpp"
 #include "ncafm.hpp"
+#include "reduce.hpp"
 #include "relaxer.hpp"
 #include "shift.hpp"
 #include "system.hpp"
@@ -37,7 +38,7 @@ MumaxWorld::MumaxWorld(real3 cellsize, Grid mastergrid, int3 pbcRepetitions)
 
 MumaxWorld::~MumaxWorld() {}
 
-void MumaxWorld::checkAddibility(Grid grid, std::string name) const {
+void MumaxWorld::checkAddibility(Grid grid, GpuBuffer<bool> geometry, std::string name) const {
   if (!inMastergrid(grid)) {
       throw std::out_of_range(
           "Can not add magnet because the grid does not fit in the "
@@ -46,7 +47,7 @@ void MumaxWorld::checkAddibility(Grid grid, std::string name) const {
 
   for (const auto& namedMagnet : magnets_) {
     Magnet* m = namedMagnet.second;
-    if (grid.overlaps(m->grid())) {
+    if (MumaxWorld::overlaps(grid, geometry, m->grid(), m->system()->geometry())) {
       throw std::out_of_range(
           "Can not add magnet because it overlaps with another "
           "magnet.");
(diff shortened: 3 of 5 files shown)


JonathanMaes (approved):

src/physics/mumaxworld.cpp
```diff
@@ -369,4 +370,36 @@ void MumaxWorld::centerDomainWall(int comp, int axis) {
...
+
+  int3 o1 = grid1.origin(), s1 = grid1.size();
+  int3 o2 = grid2.origin(), s2 = grid2.size();
+
+  // Overlapping box
+  int3 lo{std::max(o1.x, o2.x), std::max(o1.y, o2.y), std::max(o1.z, o2.z)};
+  int3 hi{std::min(o1.x + s1.x, o2.x + s2.x),
+          std::min(o1.y + s1.y, o2.y + s2.y),
+          std::min(o1.z + s1.z, o2.z + s2.z)};
+  int3 n{hi.x - lo.x, hi.y - lo.y, hi.z - lo.z};
```
lamoreel, line 399: Perhaps more meaningful variable names are appropriate here.

lamoreel: (same comment in `reduce.cu` where even less context is given.)

src/core/reduce.cu
```diff
@@ -289,4 +289,56 @@ bool isUniformField(const Field& f) {
...
+  bool found = false;
+  for (int i = tid; i < ncells; i += BLOCKDIM) {
+    // unravel flat index i into a 3D offset within [0, n)
+    int x = i % n.x;
+    int y = (i / n.x) % n.y;
+    int z = i / (n.x * n.y);
+    int3 coo{lo.x + x, lo.y + y, lo.z + z};
+
+    bool in1 = !geometry1 || geometry1[grid1.coord2index(coo)];
+    bool in2 = !geometry2 || geometry2[grid2.coord2index(coo)];
```
lamoreel, line 310: What's the reason behind the first condition `!geometry1`(2)? Why isn't the condition after the `||` sufficient?

lamoreel, line 310: If I understand correctly, `coo` is calculated from a cell within a minimal bounding box encapsulating both grids. What if `coo` is not inside a `grid`? Is this still safe?

test/test_geometry.py
```diff
@@ -52,3 +53,44 @@ def test_invalid_function_geometry(self):
...
+        plane = Cuboid(planeside, planeside, cs).translate(6*cs, 6*cs, 5*cs)
+        diff = bigCube - smallCube
+
+        magnet1 = Ferromagnet(world, grid, geometry=diff)
+        magnet2 = Ferromagnet(world, grid, geometry=plane) # or Ferromagnet(world, Grid((4, 4, 1), origin=(6, 6, 5))
+
+        assert magnet1 is not None
+        assert magnet2 is not None
+
+    def test_overlap_fail(self):
```
lamoreel, line 77: I know I gave this example a while ago as a case that should (not) be accepted as a valid geometry, but I guess in this case something like this would suffice:
```python
magnet1 = Ferromagnet(world, grid)
magnet2 = Ferromagnet(world, grid, geometry=np.zeros(grid.shape)) # or np.ones()
```
Perhaps it's also meaningful to write a test where the magnets only overlap at one corner (by shifting the origin), as to isolate the new functionality.

lamoreel (changes requested):
Glad to see my suggestion from a while back into a PR 😊 !

src/core/reduce.cu
JonathanMaes: I believe OR-statements are executed from left to right.
So if `geometry1` does not exist (or in any way results in `!geometry1` = `true`), the second half of the OR-statement is not evaluated, which in that case would lead to an indexing error.

JonathanMaes: I think it's the intersection of both grids (see `mumaxworld.cpp` line 394+), so `coo` should always be inside both grids.

lamoreel: My point exactly 😊 . The second half should be sufficient, since a global geometry check is performed in `mumaxworld.cpp` line 386 onward.

src/physics/mumaxworld.hpp
```diff
@@ -38,7 +38,7 @@ class MumaxWorld : public World {
   /** Uniform bias magnetic field which will affect all magnets in the world. */
   real3 biasMagneticField;
 
-  void checkAddibility(Grid grid, std::string name) const;
+  void checkAddibility(Grid grid, GpuBuffer<bool> geometry, std::string name) const;
```
ilateur: I think it is best here to pass geometry by constant reference `const GpuBuffer<bool>& geometry1`. Then it's not copied and unmodified.

src/physics/mumaxworld.cpp
```diff
@@ -369,4 +370,40 @@ void MumaxWorld::centerDomainWall(int comp, int axis) {
...
+  int3 overlapMaximum{std::min(o1.x + s1.x, o2.x + s2.x),
+          std::min(o1.y + s1.y, o2.y + s2.y),
+          std::min(o1.z + s1.z, o2.z + s2.z)};
+  int3 overlapCells{overlapMaximum.x - overlapOrigin.x,
+         overlapMaximum.y - overlapOrigin.y,
+         overlapMaximum.z - overlapOrigin.z};
+
+  // Check all cells in the overlapping box on GPU
+  return geometriesOverlap(grid1, hasGeo1 ? geometry1.get() : nullptr,
+                           grid2, hasGeo2 ? geometry2.get() : nullptr,
```
ilateur: I think `GpuBuffer` already returns `nullptr` if `size` is 0, so these ternary operators are unnecessary.

src/physics/mumaxworld.cpp
```diff
@@ -369,4 +370,40 @@ void MumaxWorld::centerDomainWall(int comp, int axis) {
...
+                           overlapOrigin, overlapCells);
```
ilateur: I think it would be cleaner to combine `overlapOrigin` and `overlapCells` into an `overlapGrid` and pass that as an argument.

src/core/reduce.cu
```diff
@@ -289,4 +289,59 @@ bool isUniformField(const Field& f) {
...
+}
+
+bool geometriesOverlap(Grid grid1, bool const* geometry1,
+                       Grid grid2, bool const* geometry2,
+                       int3 overlapOrigin, int3 overlapCells) {
+  int ncells = overlapCells.x * overlapCells.y * overlapCells.z;
+
+  GpuBuffer<bool> d_result(1);
+  cudaLaunchReductionKernel(k_geometriesOverlap, d_result.get(),
+                            grid1, geometry1, grid2, geometry2, overlapOrigin, overlapCells, ncells);
```
ilateur: If `overlapOrigin` and `overlapCells` are combined into an `overlapGrid` then `ncells` can be asked from that grid, either here or inside the reduction kernel.

src/core/reduce.cu
```diff
@@ -289,4 +289,59 @@ bool isUniformField(const Field& f) {
...
+  __shared__ bool sdata[BLOCKDIM];
+  int tid = threadIdx.x;
+
+  bool found = false;
+  for (int i = tid; i < ncells; i += BLOCKDIM) {
+    // unravel flat index i into a 3D offset within [0, n)
+    int x = i % overlapCells.x;
+    int y = (i / overlapCells.x) % overlapCells.y;
+    int z = i / (overlapCells.x * overlapCells.y);
+    int3 coo{overlapOrigin.x + x, overlapOrigin.y + y, overlapOrigin.z + z};
```
ilateur: `Grid` already has `index2coord` doing exactly this.

src/core/reduce.hpp
```diff
@@ -13,4 +14,7 @@ real fieldComponentAverage(const Field&, int comp);
 std::vector<real> fieldAverage(const Field&);
 bool idxInRegions(GpuBuffer<unsigned int>, unsigned int idx);
 bool isUniformFieldComponent(const Field&, int);
-bool isUniformField(const Field&);
\ No newline at end of file
+bool isUniformField(const Field&);
+bool geometriesOverlap(Grid grid1, bool const* geometry1,
+                       Grid grid2, bool const* geometry2,
+                       int3 lo, int3 n);
```
ilateur: `lo` and `n` are not very descriptive (but is also solved by passing overlap grid)

src/core/reduce.cu
ilateur: Convention in other reduction kernels seems to be to get `ncells` inside the kernel rather than passing it.

test/test_geometry.py
```diff
@@ -1,5 +1,6 @@
 import numpy as np
 import pytest
+from mumaxplus.util.shape import Cuboid
```
ilateur: `Cuboid` is never used? (sadly)

test/test_geometry.py
ilateur, line 1: It would be interesting to test is less trivial case too, for example two ellipses/ellipsoids that don't touch, but their grids do overlap.

ilateur (changes requested):

src/physics/mumaxworld.cpp
```diff
@@ -24,6 +24,8 @@
 #include "torque.hpp"
 #include "window.hpp"
 
+#include <iostream>
```
ilateur: why?

ilateur (changes requested):

src/physics/mumaxworld.cpp
DiegoDeGusem: oops

ilateur (approved):

PR #173, opened 2026-08-18
Remove duplicate strain calculation

DiegoDeGusem:
This PR removes the extra strain calculation in `evalElasticEnergyDensity` by having an optional strain parameter in `evalStressTensor` and `evalElasticStress`.


ilateur (approved):
I have quickly tested it in a 256x256x1 system and it was indeed about 5% faster 🎉

PR #174, opened 2026-08-21
Use shared step sizes for all ferromagnets in minimizer (and clean-up)

ilateur:
This PR has multiple small changes in the same couple of files. The biggest change is the shared step size. Most of my time was spent testing.

# Shared step size
The original [paper of Exl et al.](https://doi.org/10.1063/1.4862839) uses one "uniform" scalar step size for all magnetizations in a ferromagnet. I wondered if it would be better to share the step size between different linked ferromagnetic instances as well. There are two possibilities:
- ferromagnets linked by stray fields
- ferromagnets as sublattices

For the first case, I tested the MFM example, which has 2 ferromagnets in a plane. The minimization time went from 7.224 s to 8.294 s or about 15% slower for the same final state.

I also tested the multilayer example, which has 4 ferromagnets in a plane. It now takes 4.098 s instead of 3.520 s, or about 20% slower, ***but*** the final state has a lower energy. (Any comment I made about instability before was a bug in my benchmarking script.)

Without shared step:
tweaks_multilayer
With shared step:
shared-step_multilayer
So *in this particular case* a shared step is better (even though it is slower).

For the sublattices I first tested the altermagnetic Bloch wall example. The final state is the same, but the minimization time went from 1.813 s to 0.2836 s or 85% less! I also tested an altermagnetic skyrmion, which went from 12.97 s to 2.62 s or 80% less time for the same final state.

This felt like enough evidence in favor of this change.

I also changed all default `nsamples` to 10 everywhere, but changed the convergence check as well. So instead of checking 2 sublattices a total of 20 times, it checks everything together 10 times.

# Clean-up and speed-up
I also fixed some typos and clarified some mysteries with comments. Most importantly, I borrowed some tricks from mumax3 to speed up the code slightly. All previous "before-times" were with this speed-up already in place.
- Reuse `m0` and `t0` as `dm` and `dt`. This eliminates 2 field creations.
- Also use `add(dm, ...)` instead of `dm = add(...)`. This eliminates 1 field creation.
- Only initialize `m1` once.
I tested these with standard problem 2, which minimizes a variety of system sizes. Unchanged is blue, the first 2 changes are orange and all changes together are green. I did not test the 'shared step' version here, but it should not perform meaningfully different here.

compare_bench_stdp2

These changes make it up to 10% faster for small systems and about 1% faster for large systems. They do not alter the final state.

### DIFF
--- a/src/physics/minimizer.cu
+++ b/src/physics/minimizer.cu
@@ -87,6 +87,7 @@ __global__ void k_step(CuField mField,
   real3 m0 = m0Field.vectorAt(idx);
   real3 t = torqueField.vectorAt(idx);
 
+  // The explicit form of the implicit iteration scheme of eq. (8) in Exl et al.
   real t2 = dt * dt * dot(t, t);
   real3 m = ((4 - t2) * m0 + 4 * dt * t) / (4 + t2);
 
@@ -132,7 +133,9 @@ void Minimizer::step() {
 
   for (size_t i = 0; i < magnets_.size(); i++) {
     Field dm = add(real(+1), m1[i], real(-1), m0[i]);
-    Field dt = add(real(-1), t1[i], real(+1), t0[i]);  // TODO: check sign difference
+    Field dt = add(real(-1), t1[i], real(+1), t0[i]);  // opposite sign
+    // The Barzilai-Borwein step uses the difference in steepest ascend,
+    // while relax torque is the steepest *descend* direction.
 
     stepsizes_[i] = BarzilaiBorweinStepSize(dm, dt, nsteps_);
 
--- a/src/physics/minimizer.cu
+++ b/src/physics/minimizer.cu
@@ -93,7 +93,7 @@ __global__ void k_step(CuField mField,
   mField.setVectorInCell(idx, m);
 }
 
-static inline real BarzilianBorweinStepSize(Field& dm, Field& dtorque, int n) {
+static inline real BarzilaiBorweinStepSize(Field& dm, Field& dtorque, int n) {
   real nom, div;
   if (n % 2 == 0) {
     nom = dotSum(dm, dm);
@@ -134,7 +134,7 @@ void Minimizer::step() {
     Field dm = add(real(+1), m1[i], real(-1), m0[i]);
     Field dt = add(real(-1), t1[i], real(+1), t0[i]);  // TODO: check sign difference
 
-    stepsizes_[i] = BarzilianBorweinStepSize(dm, dt, nsteps_);
+    stepsizes_[i] = BarzilaiBorweinStepSize(dm, dt, nsteps_);
 
     addMagDiff(maxVecNorm(dm));
   }
--- a/src/physics/minimizer.cu
+++ b/src/physics/minimizer.cu
@@ -131,8 +131,10 @@ void Minimizer::step() {
     t1[i] = torques_[i].eval();
 
   for (size_t i = 0; i < magnets_.size(); i++) {
-    Field dm = add(real(+1), m1[i], real(-1), m0[i]);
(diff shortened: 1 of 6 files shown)


jplauzie:
Wow, that's pretty neat. I would've naively expected separate stepsizes to work better, so 'stiffer' magnets could advance separately. But I guess this better forces it as a global minimization problem?

It looks like you're just summing the dmdm/dmdt terms for each magnet, as $s = \frac{\sum_i \Delta m_i \cdot \Delta m_i}{\sum_i \Delta m_i \cdot \Delta t_i}$. Does it make any difference if you just take the minimum $Min\left( \frac{\Delta m_i \cdot \Delta m_i}{\Delta m_i \cdot \Delta t_i} \right)$ (might be super slow though... it'd force each one to evolve at the pace of the smallest stepping magnet)?

ilateur:
Hi @jplauzie!

> I would've naively expected separate stepsizes to work better, so 'stiffer' magnets could advance separately. But I guess this better forces it as a global minimization problem?

I would also have expected that evolving each magnet separately would allow each magnet to minimize as quickly as possible, but it seems not to be the case for "tightly linked" systems such as altermagnets.

I have only timed it and have not looked into the minimization path, so this is purely speculative, but perhaps evolving each sublattice with a different step can create oscillating states. Each sublattice could adjust mainly due to the strong coupling to the other, then take a different individual step and end up needing to adjust strongly again to the differently evolved sublattice. This could cause the system to ping-pong around the global minimum for a while (again, this is speculation).

Of course, if two systems are completely uncoupled, their global minimum should be found fastest by minimizing each separately. Here, the ferromagnets coupled only via stray fields seem weakly coupled enough that evolving them individually is still faster. There must be some point in between where this switches.

> It looks like you're just summing the dmdm/dmdt terms for each magnet, as s = ∑ i Δ m i ⋅ Δ m i ∑ i Δ m i ⋅ Δ t i . Does it make any difference if you just take the minimum M i n ( Δ m i ⋅ Δ m i Δ m i ⋅ Δ t i ) (might be super slow though... it'd force each one to evolve at the pace of the smallest stepping magnet)?

I am indeed summing the numerator and denominator separately before calculating the step. This seemed like the natural expansion of `dotSum` (or the scalar product in the [original paper by Barzilai and Borwein](https://doi.org/10.1093/imanum/8.1.141)). For example, splitting a ferromagnet into left and right halves, we would have $\langle \Delta {\bf m}, \Delta {\bf m} \rangle = \langle \Delta {\bf m}_l, \Delta {\bf m}_l \rangle + \langle \Delta {\bf m}_r, \Delta {\bf m}_r \rangle$ and similarly for $\Delta {\bf t}$. This should therefore also approximately solve the secant equation for the whole system. I would expect this to be faster than taking the minimum of the two steps, although I have not tested that yet.

That said, this minimum operation may be useful in light of #171 to merge magnetic and elastic step sizes, since the addition $\langle \Delta {\bf m}, \Delta {\bf m} \rangle + \langle \Delta {\bf u}, \Delta {\bf u} \rangle$ does not make sense. The resulting step sizes also have different units, though perhaps they could be rescaled? Or perhaps a completely different minimization approach is necessary... But this is not the place for that discussion.

DiegoDeGusem (approved):

src/physics/minimizer.cu
```diff
@@ -64,8 +55,15 @@ Minimizer::Minimizer(const MumaxWorld* world,
 
   for (auto magnet : magnets_)
     torques_.push_back(relaxTorqueQuantity(magnet));
-    
-  stepsizes_.assign(N, 1e-14);
+
+  nMagDiffSamples_ = nMagDiffSamples;
```
lamoreel, line 59: This can now be moved to the member initialization list before the constructor body.

lamoreel (approved):

PR #178, opened 2026-08-31
Proposed standard problem of DW pinning

ilateur:
About a year ago I added the proposed standard problem of domain wall pinning of the [paper by Heistracher et al.](https://doi.org/10.1016/j.jmmm.2021.168875) as an example. For some reason I intended to convert this to a test but I never did. So I dusted off the old branch and added the test.

This standard problem is designed to test the exchange strength between different regions. If done incorrectly, the pinning field would be much smaller.


examples/standardproblem_DW_pinning.py
```diff
@@ -0,0 +1,88 @@
...
+
+# material parameters
+Asoft = 0.25e-11
+Ahard = 1.0e-11
+Ksoft = 1e5
+Khard = 1e6
+Ms_soft = 0.25 / MU0
+Ms_hard = 1 / MU0
+
+# choose combination of soft and hard parameters for phase 1
```
DiegoDeGusem: and phase 2? (same comment in other files)

JonathanMaes (approved):

DiegoDeGusem (approved):

PR #183, opened 2026-09-16
Add initial time step to various timesolver methods

ilateur:
The `Timesolver` methods `steps`, `run`, `run_while` and `solve` set a sensible time step before running the solver. This could overwrite your own time step when adaptive time stepping is turned on. Sometimes (mostly for debugging) you want to use your own initial guess and the only way to circumvent this was to do the following.
```python
world.timesolver.timestep = your_initial_timestep
world.timesolver._impl.run(duration)  # this would not set sensible timestep
```
I have personally needed to use this and have recommended this before, but this is not intuitive or clear behavior at all. This pull request adds `initial_timestep` as an optional argument to these functions, as well as some simple tests.


DiegoDeGusem (approved):

mumaxplus/timesolver.py
```diff
@@ -106,35 +106,82 @@ def set_method(self, method_name):
...
+        Parameters
+        ----------
+        nsteps : int
+            The number of steps to take.
+        initial_timestep : float, optional
+            The time step is set to this value before running the solver.
+            If None (default), the initial time step is set to a sensible value
+            if adaptive timestepping is enabled, or remains unchanged if
+            disabled.
+        """
```
JonathanMaes, line 125: No "see also" like the other descriptions?

JonathanMaes (approved):

